PR #3, opened 2021-03-16
Initial RV64 preparations

cerg2010cerg2010:


src/clint.c
```diff
@@ -17,12 +17,13 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
...
 
 #include "clint.h"
+#include "riscv32.h"
 #include "riscv32_mmu.h"
 #include "rvtimer.h"
 #include "mem_ops.h"
 #include "bit_ops.h"
 
-bool clint_mmio_handler(riscv32_vm_state_t* vm, riscv32_mmio_device_t* device, uint32_t offset, void* data, uint32_t size, uint8_t access)
+bool clint_mmio_handler(riscv32_vm_state_t* vm, riscv32_mmio_device_t* device, physaddr_t offset, void* data, uint32_t size, uint8_t access)
```
LekKit, line 26: Change offset type to ptrdiff_t, since it is relative to MMIO device base address
Change size type to size_t
This applies to every MMIO callback

src/riscv32.h
```diff
@@ -66,9 +68,19 @@ enum
...
+enum
+{
+	ISA_RV32 = 1,
+	ISA_RV64,
+	ISA_RV128
+};
+
+#define ISA_MAX ISA_RV32
+
 #define INTERRUPT_MASK 0x80000000
```
LekKit, line 84: This will be changed in future for proper RV64 support

src/riscv32_csr_m.c
```diff
@@ -17,30 +17,51 @@ You should have received a copy of the GNU General Public License
...
 
 #define CSR_MARCHID 0x5256564D // 'RVVM'
 
-// no N extension, U_x bits are hardwired to 0
-#define CSR_MSTATUS_MASK 0x807FF9CC
 #define CSR_MEIP_MASK    0xAAA
 
 #define CSR_MISA_RV32  0x40000000
 #define CSR_MISA_RV64  0x80000000
 #define CSR_MISA_RV128 0xC0000000
```
LekKit, line 30: Needs to be shifted for RV64 as well

src/riscv32_csr_s.c
```diff
@@ -17,84 +17,111 @@ You should have received a copy of the GNU General Public License
...
-    uint32_t satp = (vm->mmu_virtual ? 0x80000000 : 0) | (vm->root_page_table >> 12);
+    reg_t satp;
+    bool is32bit = vm->isa[PRIVILEGE_SUPERVISOR] == ISA_RV32;
+    if (is32bit)
+    {
+	    satp = (!!vm->mmu_virtual << 31) | (vm->root_page_table >> 12);
+    }
+    else
+    {
+	    satp = ((reg_t)vm->mmu_virtual << 60) | ((vm->root_page_table >> 12) & gen_mask(44));
```
LekKit, line 110: GCC: warning: left shift count >= width of type
Probably should be separated with ifdef

src/riscv32i.c
```diff
@@ -119,16 +119,27 @@ static void riscv32i_srli_srai(riscv32_vm_state_t *vm, const uint32_t instructio
...
+    size_t extend_from = XLEN(vm);
+    for (size_t i = 5; i < XLEN_BIT(vm); ++i)
+    {
+	    if (!is_bit_set(shamt, i))
+	    {
+		    extend_from = 1 << i;
+		    src_reg &= gen_mask(extend_from);
+		    break;
+	    }
+    }
```
LekKit, line 136: This should not be done here, separate instruction implementations with separate dispatch tables are going to be used for RV64. New srli/srai/slli are 3-4 times slower than the original implementation, and this is likely to be redundant. I suggest reverting this change.

src/riscv32i.c
```diff
@@ -119,16 +119,27 @@ static void riscv32i_srli_srai(riscv32_vm_state_t *vm, const uint32_t instructio
...
 
-    if (funct7 == 0x20) {
-        riscv32i_write_register_s(vm, rds, ((int32_t)src_reg) >> shamt);
+    if (funct7 == 1 << (12 - XLEN_BIT(vm) - 2)) {
```
LekKit: GCC: warning: comparison of integer expressions of different signedness

src/riscv32_mmu.c
```diff
@@ -103,60 +148,144 @@ static void tlb_put(riscv32_vm_state_t* vm, uint32_t addr, uint32_t page_addr, u
 }
 
 // Virtual memory addressing mode (SV32)
-static bool riscv32_mmu_translate_sv32(riscv32_vm_state_t* vm, uint32_t addr, uint8_t access, uint32_t* dest_addr)
+bool riscv_mmu_translate(riscv32_vm_state_t *vm, const virtaddr_t va, physaddr_t *pa, const uint8_t access, bool update_pages)
```
LekKit: The soft MMU is a lot slower now, Linux system boot takes twice as much time, ramspeed reports similar regressions in raw memory speed. I think SV32 & SV39/48 translations should be separate functions as intended previously, and more optimally written, otherwise the compiler optimizations are less efficient.

src/main.c
```diff
@@ -111,60 +113,72 @@ int main(int argc, char** argv)
...
-	    uint32_t dts = load_file_to_ram(vm, dtb_addr, args.dtb);
+	    physaddr_t dts = load_file_to_ram(vm, dtb_addr, args.dtb);
 	    if (dts == 0)
 	    {
 		    printf("ERROR: Failed to load DTB.\n");
 		    return 1;
 	    }
 
-	    printf("DTB loaded at: 0x%zx size: %"PRId32"\n", dtb_addr, dts);
+	    printf("DTB loaded at: 0x%"PRIxpaddr" size: %"PRIxpaddr"\n", dtb_addr, dts);
```
LekKit, line 151: Size is shown as hex without 0x prefix, not very straightforward - looks like incorrect decimal size

src/riscv32_csr_m.c
cerg2010cerg2010: It is, that's a removal. This constant is now replaced with csr_mstatus_mask function

LekKit:
For anyone reading this in future:
We have personally discussed about slightly different RV64 implementation approach, and this code is going to be partially reused.
As for now, closing this PR.

PR #20, opened 2021-06-09
F & D extensions implementation, some fixes

cerg2010cerg2010:


LekKit:
👍
Some future notes:
- We need to move the decoder to the hart struct (planned with API rewrite)
- Cleanly disable/ignore host FPU exceptions
- Canonical NaN for JIT(?)

PR #22, opened 2021-07-27
FDT library, USE_FPU flag, VMSWAP mechanism, usual fixes

cerg2010cerg2010:


LekKit (approved):
Nice!
As for split vmswap, i would suggest using paddr-based filenames instead, but this is not urgent
 :+1:

PR #36, opened 2021-09-25
Tiny FPU refactoring and endless fixes

cerg2010cerg2010:
LGTM warnings should be gone.


LekKit:
Hmm, seems to reduce performance by ~8% in Dhrystone, and by 5% in Whetstone. Coremark results are unaffected.
I assume this could be optimized in future, and leave it FYI

LekKit (approved):

PR #46, opened 2022-01-28
Basic GitHub CI support

000exploit:
Currently, builds only x86_64 binary with RV64, JIT, FPU and PCI for Linux and MacOS, may be useful in the near future.


flrdv:
why not??

Mr0maks:
> why not??

Gitlab CI is better

000exploit:
> why not??

There's no ARM64 build support, no binary stripping and the CI script is also a bit messed up. Maybe I'll rewrite it in the future.

LekKit:
More refined version implemented in ffc0495. Thanks for participation, I assume we can close this PR.

PR #50, opened 2022-03-31
Implement sparse block format with deduplication, trim

0xCatPKG:
Implements sparse block format named Risc-V Virtual Drive
- [ ]  Implement basic functions
- [ ] Implement "overlay" mode
- [ ] Implement format conversion (from .img, to .img)
- [ ] Implement trim
- [ ] Implement deduplication
- [ ] Rewrite for async io support
- [ ] Implement blk_dev support


LekKit:
Possible suggestions (rev.1) for RVVD format:
- Virtual->physical block lookup table, lazy allocation
- Reverse lookup table from physical blocks to virtual, refcounts (For efficient TRIM)
- Open-addressing hashtable from block hashes to their physical location (For deduplication)
- Copy on Write: Each newly written block either dedups against existing one (simply +1 reference) or is written into unallocated space, the references are updated afterwards
- This brings us ~24-32 bytes of metadata (In different places) per each possible physical block in the image. On-demand metadata allocation schemes would be good to investigate.
- Use fast & modern hash algorithm like [XXH3](https://github.com/Cyan4973/xxHash). It has amazing performance/distribution characteristics, collision resistance isn't needed here
- Compression doesn't sound very cool: It brings fragmentation & variable block sizes, breaks alignment (Which could cause read/write amplification)

This of course is only a lists of suggestions from my ideas, another possible design is to use BTRFS-like extents (Which have variable size in blocks). Such design is particularly good for compression, but suffers from fragmentation & most likely will have issues with dedup/TRIM. In my opinion, this is not well suited for block devices in virtual machines.

0xCatPKG:
Since I am no longer part of the RVVM community, I am closing this PR

PR #51, opened 2022-04-01
Implement blk_dev API and own file API

0xCatPKG:
- [x] Linux support
- [x] Windows support
- [x] ~~Implement aio support in IO Thread~~
- [x] ATA first step refactoring
- [x] Rewrite ATA using blk_dev API
- [x] Check OSX & FreeBSD support
- [x] Write documentation to new API's


LekKit:
@0xCatPKG, should we remove the Linux AIO implementation?
io_submit() seems to block sometimes, so there is little rationale as opposed to IO threads

LekKit (approved):

PR #62, opened 2022-12-03
Fix argument to ReadConsoleW

j-mattsson:
(I have no experience with this code base so someone should probably review this before merging.)

### DIFF
--- a/src/devices/ns16550a.c
+++ b/src/devices/ns16550a.c
@@ -137,7 +137,7 @@ static uint8_t terminal_readchar(void* addr)
     if (t_head == t_tail && _kbhit()) {
         wchar_t w_buff[8];
         DWORD w_chars;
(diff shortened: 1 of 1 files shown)


LekKit (approved):
Yeah I approve this, thanks. Somehow I missed that `ReadConsoleW` takes number of characters instead of buffer size, and the lack of continuous Win32 testing let it like that (Tho it's weird that none of our static analyzers warned on it).
All builds (MinGW, MSVC, mingwCE) passed. Will test on a Windows VM soon before merging.

I assume you caught this by copy-pasting a lot of text sorta?

LekKit:
Tested that VT works properly on Win32, no problems with large pasted inputs.
It seems that static analyzers missed it because the project is mostly developed/tested/analyzed on Linux, whether Win32-specific code doesn't receive that much attention. This is a sign this needs improvement in future.

image

PR #63, opened 2022-12-04
haiku_window: hide cursor, add capture on mouse down

X547:


LekKit:
I assume `B_POINTER_EVENTS` is used to capture pointer movement when it is outside the window? Similar behavior is default on X11. Doesn't it need to be reset afterwards?

Somewhat related: When I press the mouse and move it beyond top window border, guest cursor jumps to screen bottom on X. Does this also happen on Haiku or is it something else?

X547:
> I assume `B_POINTER_EVENTS` is used to capture pointer movement when it is outside the window? Similar behavior is default on X11. Doesn't it need to be reset afterwards?

`BView::SetMouseEventMask` API cause to receive pointer events even when pointer is outside BView/BWindow bounds. It will automatically reset on release. There is also `BView::SetEventMask` API that do not automatically reset on release and allow keyboard/mouse event monitoring.

> Somewhat related: When I press the mouse and move it beyond top window border, guest cursor jumps to screen bottom on X. Does this also happen on Haiku or is it something else?

Do you use latest patches versions? I added range check in HID mouse implementation.

PR #67, opened 2022-12-05
i2c-hid: update

### LINKED ISSUE #58: Absolute positioned input device support
Currently RVVM support only relative-positioned PS/2 mouse emulation that require pointer capture. It is annoying and limiting because it do not work properly on touchscreens and tablets with digital pen. TinyEMU and QEmu supports absolute positioned pointer.

It can be implemented with i2c HID or USB HID.

X547:


LekKit:
Will test this for a bit to make sure we didn't miss anything this time, thanks

LekKit (approved):

PR #74, opened 2023-01-02
Add more detailed error codes to the `rvvm_dump_dtb` API function

ghost:
This introduces 3 possible errors:
- `DTB_DUMP_ERROR_OPEN`: indicates that operation was failed due to failure in `rvopen`
- `DTB_DUMP_ERROR_NO_FDT`: indicates that operation was failed due to lack of `FDT` feature
- `DTB_DUMP_OK`: indicates that operation was successfully performed

This change is not actually breaking for the RVVM codebase, but possibly breaking for someone's else codebase if rvvmlib is used in it with the smth like:
```C
if (rvvm_dump_dtb("...")) {
 // Ok
} else {
 // Err
}
```

I can solve this by introducing new function with name  like `rvvm_dump_dtb_ex`, but do we really need it? Objections appreciated

### DIFF
--- a/src/rvvm.c
+++ b/src/rvvm.c
@@ -485,6 +485,10 @@ PUBLIC bool rvvm_load_dtb(rvvm_machine_t* machine, const char* path)
     return file_reopen_check_size(&machine->dtb_file, path, machine->mem.size >> 1);
 }
 
+PUBLIC bool rvvm_dump_dtb_is_ok(dtb_dump_result_t result) {
+    return result == DTB_DUMP_OK;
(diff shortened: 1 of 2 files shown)


LekKit:
This isn't quite what I'd want from an API standpoint, and USE_FDT is almost never disabled. When it's disabled it's most likely a custom minimal build (So all bets are off about the feature)

How about some kind of `rvvm_errno`?

ghost:
Yes, that sounds like solution, I'll work on it

PR #80, opened 2023-03-10
x11: Hotkey to grab input

ghost:
Add a hotkey Ctrl+Alt+G to grab keyboard and mouse input. Ungrab should be done using the same modifiers (left/right alt/ctrl) as the grab.

### DIFF
--- a/src/devices/x11window_xlib.c
+++ b/src/devices/x11window_xlib.c
@@ -29,6 +29,17 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #include <sys/shm.h>
 #endif
 
+enum key_mask {
+    KEY_MASK_LEFTCTRL = 1 << 0,
+    KEY_MASK_RIGHTCTRL = 1 << 1,
+    KEY_MASK_LEFTALT = 1 << 2,
+    KEY_MASK_RIGHTALT = 1 << 3,
+};
+
+static const char x11_title[] = "RVVM";
+
+static const char x11_title_capture[] = "RVVM - Press Ctrl+Alt+G to release mouse/keyboard";
+
 struct win_data {
     Display* display;
     Window window;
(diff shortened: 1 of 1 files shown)


LekKit:
Few things would be nice to add:
- Actually relative mouse input (Use `hid_mouse_move()` with relative coordinates from X11), so that games can get mouse input moving around indefinitely and not getting stuck
- Maybe just use any L/R button as a modifier. Is there a reason why it should unlock with left buttons if locked by them in the first place? Plus the added burden of string handling...

Overall pretty nice, I like it. I just wanna set some decisions in stone so I can later implement 1:1 behavior in other GUI implementations.

src/devices/x11window_xlib.c
```diff
@@ -199,6 +222,47 @@ static rgb_fmt_t x11_get_rgb_format(Display* display)
...
+        win->data->grab_modkeys = win->data->pressed_modkeys;
+
+        XGrabKeyboard(dsp, win->data->window, True, GrabModeAsync,
+                GrabModeAsync, CurrentTime);
+        XGrabPointer(dsp, win->data->window, True,
+                ButtonPressMask | ButtonReleaseMask | PointerMotionMask,
+                GrabModeAsync, GrabModeAsync, win->data->window, None, CurrentTime);
+
+        char buf[128], *p = buf;
+        p = stpcpy(p, "RVVM - Press ");
```
LekKit: I do have suspicion this will not be supported on some targets... Correct if I'm wrong

LekKit (commented):
Few things would be nice to add:
- Actually relative mouse input (Use `hid_mouse_move()` with relative coordinates from X11), so that games can get mouse input moving around indefinitely and not getting stuck
- Maybe just use any L/R button as a modifier. Is there a reason why it should unlock with left buttons if locked by them in the first place? Plus the added burden of string handling...

Overall pretty nice, I like it. I just wanna set some decisions in stone so I can later implement 1:1 behavior in other GUI implementations.

ghost:
> Actually relative mouse input (Use `hid_mouse_move()` with relative coordinates from X11), so that games can get mouse input moving around indefinitely and not getting stuck

Currently, `hid-mouse.c` reports absolute coordinates like a tablet. So even if I use `hid_mouse_move`, it is still bounded by screen dimensions.

I think we need to add support for reporting relative coordinates. Not sure if we should add a new hid device or extend the current one to support both absolute and relative coordinates. According to the source code, qemu seems to provide separate hid mouse and hid tablet.

> Maybe just use any L/R button as a modifier. Is there a reason why it should unlock with left buttons if locked by them in the first place? Plus the added burden of string handling...

I've mimicked qemu behavior here. I'll remove it you prefer not to have this.

PR #88, opened 2023-04-12
Makefile: Fix build on OpenBSD

ghost:
Define _POSIX_C_SOURCE to:
- Avoid paddr_t and vaddr_t from BSD
- Enable clock_gettime and CLOCK_MONOTONIC

Add /usr/X11R6 to CFLAGS and LDFLAGS

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -152,6 +152,10 @@ endif
 ifeq ($(OS),serenityos)
 USE_SDL ?= 1
 endif
+ifeq ($(OS),openbsd)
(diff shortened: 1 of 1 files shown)


LekKit:
Hmm.. `vaddr_t` types and such could collide, interesting if that should be fixed or some header isolation is needed

LekKit:
UPD: Turns out OpenBSD `vaddr_t` manifests itself though `<stdlib.h>` which *is* included generally everywhere. I did not expect someone would put system-specific definitions in C standard library, hence the clash.

It might be wise to do more header isolation to prevent such clashes from occurring in the future.

PR #89, opened 2023-04-15
Fix NS16550A on OpenBSD

ghost:
Hello, I find that when openbsd stucked at console:
- it had set plic irq thresold to 3
- rvvm notify a irq from uart with priority 3, but didn't send it.
- it then change its irq thresold between 0 and 3.

If irqs was to be scanned after thresod write, then the kept irq will be fired.
I don't know where (unable) to find the actually behaviour in plic datasheet or manual, so this change may be naive, but it does works...

Does my guess right?  Thank you!

### DIFF
--- a/src/devices/ns16550a.c
+++ b/src/devices/ns16550a.c
@@ -270,10 +270,6 @@ static void ns16550a_update(rvvm_mmio_dev_t* device)
         regs->len = regs->len ? 1 : terminal_readchar(&regs->buf);
         if (regs->len && (regs->ier & NS16550A_IER_RECV)) {
             plic_send_irq(regs->plic, regs->irq);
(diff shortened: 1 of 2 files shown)


LekKit:
Ah, so my gut feeling that this is a PLIC bug was right. Ironically, I thought that rewriting the PLIC would help (The reason why PLIC is lock-free now), but stumbled upon it anyways.

So yeah, changing the source priority should trigger the IRQ scan as well, kudos for fixing that! :D 

P.S. PLIC documentation is here: https://github.com/riscv/riscv-plic-spec/blob/master/riscv-plic.adoc

LekKit (approved):

ghost:
Cool, I got that PLIC documentation but can't say I understand it well..(

Issue #86, opened 2023-04-10
Asking for more serial ports


fish4terrisa-MSDSM:
(sorry for my broken english...)
I made a project of a emulator on android ,which runs archlinux riscv on android using RVVM.The project is [archriscv-term](https://github.com/fish4terrisa-MSDSM/archriscv-term).The problem is that there is only one tty available.Could you add more serial ports? You may bind them to some sockets so that I can connect them using socat.

LekKit:
I tried to add multiple serial ports previously, they were fine hardware-wise but the system didn't care to put kernel console or `logind` on them. Any help guest-side is appreciated.

fish4terrisa-MSDSM:
Try `systemctl enable --now getty@ttyS1.service`.It will work .(Tested in qemu)

LekKit:
Any suggestions for a nice serial device API? Doing sth similar to `tap_api` would require an internal ring buffer, hence I imagine sth like this instead:
- `serial_write()` called from uart and writes to device (TX)
- `serial_read()` called from uart and reads bytes from device (RX)
- `serial_notify()` called from device side to notify there are bytes to receive (Basically a software interrupt)

@iyzsong is working on basically same thing, see #85 (And mentioned uart sockets) so I won't touch `ns16550a` for now to prevent merge conflicts. The serial API will be reviewed considering anything we will came up with. Stay tuned X)

LekKit:
FYI: I will be posting Linux kernel configs and stuff soon (Also working on a proper RVVM U-Boot support) so you can build newer mainline kernels without copying my old builds around

fish4terrisa-MSDSM:
> FYI: I will be posting Linux kernel configs and stuff soon (Also working on a proper RVVM U-Boot support) so you can build newer mainline kernels without copying my old builds around

Hurry ,that really helps me .Its hard to recompile the mainline kernel of arch.(It has took me more than 1 day to compile in archriscv-term with rvvm)

LekKit:
> Its hard to recompile the mainline kernel of arch.(It has took me more than 1 day to compile in archriscv-term with rvvm)

See https://github.com/LekKit/patches-misc for a minimal kernel config, U-Boot patch, etc useful things. I haven't decided where to post binary builds yet.

Use a `riscv64-linux-gnu` crosscompiler toolchain natively on your computer, compiling in a VM is going to be several times slower. It is available on many distributions including Arch or Debian-based distros.
[...]

PR #90, opened 2023-04-16
ns16550a: Support different backends

ghost:
WIP for preview and early review if interested, thank you!

- introduce `chardev_t` as ns16550a backend.  OO pattern from https://nullprogram.com/blog/2014/10/21/
- refactorying out terminal backend as `ns16550a_term.c`
- add `ns16550a_tcp.c` backend for testing

Connect as `socat stdio,echo=0,raw tcp:localhost:3000`, multiple clients supported.
Need add a cmd option to select backend, more error handling and cleanup...

### DIFF
--- a/src/devices/chardev.h
+++ b/src/devices/chardev.h
@@ -1,0 +1,56 @@
+/*
+chardev.h - backend for UART
+Copyright (C) 2023  宋文武 <iyzsong@envs.net>
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#ifndef CHARDEV_H
+#define CHARDEV_H
+
+#include "rvvmlib.h"
+
+typedef struct chardev chardev_t;
+struct chardev {
+    size_t (*read)(chardev_t* dev, void* buf, size_t nbytes);
+    size_t (*write)(chardev_t* dev, const void* buf, size_t nbytes);
+    void (*destroy)(chardev_t* dev);
+    void (*watch)(chardev_t* dev,
+                  void (*on_input_available)(chardev_t* dev, void* watcher_data),
+                  void (*on_output_available)(chardev_t* dev, void* watcher_data),
+                  void* watcher_data);
+    void* data;
+};
+
+static inline size_t chardev_read(chardev_t* dev, void* buf, size_t nbytes)
+{
+    return dev->read(dev, buf, nbytes);
+}
+static inline size_t chardev_write(chardev_t* dev, const void* buf, size_t nbytes)
+{
+    return dev->write(dev, buf, nbytes);
+}
+static inline void chardev_watch(chardev_t* dev,
+                                 void (*on_input_available)(chardev_t* dev, void* watcher_data),
+                                 void (*on_output_available)(chardev_t* dev, void* watcher_data),
+                                 void* watcher_data)
+{
+    return dev->watch(dev, on_input_available, on_output_available, watcher_data);
+}
+static inline void chardev_destroy(chardev_t* dev)
+{
+    return dev->destroy(dev);
+}
+
+#endif
--- a/src/devices/chardev_term.c
+++ b/src/devices/chardev_term.c
@@ -1,0 +1,376 @@
+/*
+chardev_term.c - Terminal backend for UART
+Copyright (C) 2021  LekKit <github.com/LekKit>
+                    Mr0maks <mr.maks0443@gmail.com>
+Copyright (C) 2023  宋文武 <iyzsong@envs.net>
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#include "chardev_term.h"
+#include "threading.h"
+#include "utils.h"
+
+#if (defined(__unix__) || defined(__APPLE__) || defined(__HAIKU__)) && !defined(__EMSCRIPTEN__)
+#include <sys/types.h>
+#include <sys/select.h>
+#include <termios.h>
+#include <unistd.h>
+#include <string.h>
+#include <errno.h>
+
+
+static const uint8_t TERM_CTL_WAIT_READ  = 1;
+static const uint8_t TERM_CTL_WAIT_WRITE = 2;
+static const uint8_t TERM_CTL_STOP       = 3;
+static const uint8_t TERM_CTL_CONT       = 4;
+typedef struct {
+    thread_ctx_t* thread;
+    int ctl[2];
+    void (*on_input_available)(chardev_t* dev, void* watcher_data);
+    void (*on_output_available)(chardev_t* dev, void* watcher_data);
+    void* watcher_data;
+} chardev_term_t;
+
+static struct termios orig_term_opts;
+
+static void terminal_origmode()
+{
+    tcsetattr(STDIN_FILENO, TCSAFLUSH, &orig_term_opts);
+}
+
+static void terminal_rawmode()
+{
+    tcgetattr(STDIN_FILENO, &orig_term_opts);
+    atexit(terminal_origmode);
+    struct termios term_opts = orig_term_opts;
+    term_opts.c_lflag &= ~(ECHO | ICANON | ISIG | IEXTEN);
+    term_opts.c_iflag &= ~(IXON | ICRNL);
+    tcsetattr(STDIN_FILENO, TCSAFLUSH, &term_opts);
+}
+
+static void* terminal_thread(void* arg)
+{
+    chardev_t* dev = (chardev_t*)arg;
+    chardev_term_t* term = (chardev_term_t*)dev->data;
+    int ctlfd = term->ctl[0];
+    bool watch_read = false;
+    bool watch_write = false;
+
+    while (true) {
+        fd_set rfds, wfds;
+        FD_ZERO(&rfds);
+        FD_ZERO(&wfds);
+        FD_SET(ctlfd, &rfds);
+        if (watch_read)
+            FD_SET(0, &rfds);
+        if (watch_write)
+            FD_SET(1, &wfds);
+        if (select(ctlfd + 1, &rfds, &wfds, NULL, NULL) == -1) {
+            rvvm_error("select() failed: %s", strerror(errno));
+            break;
+        }
+        if (FD_ISSET(ctlfd, &rfds)) {
+            uint8_t ctl = 0;
+            if (read(ctlfd, &ctl, 1) < 1)
+                break;
+            if (ctl == TERM_CTL_STOP) {
+                // Wait 'terminal_watch' to update callbacks.
+                while (ctl != TERM_CTL_CONT) {
+                    if (read(ctlfd, &ctl, 1) != 1)
+                        return NULL;
+                }
+                if (term->on_input_available)
+                    watch_read = true;
+                if (term->on_output_available)
+                    watch_write = true;
+            }
+            if (ctl == TERM_CTL_WAIT_READ && term->on_input_available)
+                watch_read = true;
+            if (ctl == TERM_CTL_WAIT_WRITE && term->on_output_available)
+                watch_write = true;
+        }
+        if (FD_ISSET(0, &rfds) && term->on_input_available) {
+            watch_read = false;
+            term->on_input_available(dev, term->watcher_data);
+        }
+        if (FD_ISSET(1, &wfds) && term->on_output_available) {
+            watch_write = false;
+            term->on_output_available(dev, term->watcher_data);
+        }
+    }
+    return NULL;
+}
+
+static void terminal_watch(chardev_t* dev,
+                           void (*on_input_available)(chardev_t* dev, void* watcher_data),
+                           void (*on_output_available)(chardev_t* dev, void* watcher_data),
+                           void* watcher_data)
+{
+    chardev_term_t* term = (chardev_term_t*)dev->data;
+    if (write(term->ctl[1], &TERM_CTL_STOP, 1) != 1)
+        close(term->ctl[1]);
+    term->on_input_available = on_input_available;
+    term->on_output_available = on_output_available;
+    term->watcher_data = watcher_data;
+    if (write(term->ctl[1], &TERM_CTL_CONT, 1) != 1)
+        close(term->ctl[1]);
+}
+
+static void terminal_destroy(chardev_t* dev)
+{
+    chardev_term_t* term = (chardev_term_t*)dev->data;
+    if (term->thread == NULL)
+        return;
+    close(term->ctl[1]);
+    thread_join(term->thread);
+    close(term->ctl[0]);
+    term->thread = NULL;
+}
+
+static size_t terminal_read(chardev_t* dev, void* buf, size_t nbytes)
+{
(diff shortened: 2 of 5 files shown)


LekKit:
Generally nicely designed but a few caveats:
- You shouldn't include `<unistd.h>` or any OS-specific header outside of `networking.c`/`threading.c`/`blk_io.c`/etc wrappers. I know you wanted to use `ssize_t`, I think it is possible to use `size_t` unless you want a very detailed error handling (Usually there is no point esp. since different backends will be different, and in general it is nice when there is no error other than host-induced which is immediately reported via `rvvm_error()`
- Some build issues in different places, I hope you can resolve them (Or invite me in your repo)
- You are apparently relying on the fact that you can cast `chardev_t*` into pointer to different struct that holds `chardev_t` as a first member. This is a non-portable hack and might break on weird targets like Emscripten. You are better of adding some opaque `void*` pointer as `chardev_t` field and casting it into a pointer to `*some backend-defined struct*`
- Rename other term implementations to `chardev_term.c`, `chardev_tcp.c` because they have nothing to do with NS16550 (There are other types of uarts etc etc)

There are things that need some questions and possible redesign:
- Earlier polling implementation is possibly sub-ideal and I'd want a way to call back into UART and explicitly send an interrupt, because I'll be running a separate terminal thread for proper event-driven approach. This asks for a bit of redesign to `chardev_t` api
- Shouldn't different TCP connections work with different UARTs? Or, there should be multiple ports per each UART instead?

ghost:
> Generally nicely designed but a few caveats:

Thanks, this is very help for me!

 
> * You shouldn't include `<unistd.h>` or any OS-specific header outside of `networking.c`/`threading.c`/`blk_io.c`/etc wrappers. I know you wanted to use `ssize_t`, I think it is possible to use `size_t` unless you want a very detailed error handling (Usually there is no point esp. since different backends will be different, and in general it is nice when there is no error other than host-induced which is immediately reported via `rvvm_error()`

Okay, make sense.


> * Some build issues in different places, I hope you can resolve them (Or invite me in your repo)

some `USE_NET` conditions, yeah I can fix it later..

> * You are apparently relying on the fact that you can cast `chardev_t*` into pointer to different struct that holds `chardev_t` as a first member. This is a non-portable hack and might break on weird targets like Emscripten. You are better of adding some opaque `void*` pointer as `chardev_t` field and casting it into a pointer to `*some backend-defined struct*`

Well, I find this in c99 spec https://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf   6.7.2.1 13:
```
A pointer to a structure object, suitably converted, points to its initial member (or if that member is a
bit-field, then to the unit in which it resides), and vice versa. There may be unnamed
padding within a structure object, but not at its beginning.
```
Which should be safe IIUC, but I don't know about Emscripten, so I modify thoes to use a `void* userdata` pointer.


> * Rename other term implementations to `chardev_term.c`, `chardev_tcp.c` because they have nothing to do with NS16550 (There are other types of uarts etc etc)

Okay.

> There are things that need some questions and possible redesign:
> 
> * Earlier polling implementation is possibly sub-ideal and I'd want a way to call back into UART and explicitly send an interrupt, because I'll be running a separate terminal thread for proper event-driven approach. This asks for a bit of redesign to `chardev_t` api

I'm not sure how to implement it, does every `chardev_t` has its thread looping with select/poll and call a callback from uart?

> * Shouldn't different TCP connections work with different UARTs? Or, there should be multiple ports per each UART instead?

Multiple ports should works.

I also add a `chardev_mux` to combine two chardev_t as one, so term and tcp can show a uart session sync..

ghost:
> Earlier polling implementation is possibly sub-ideal and I'd want a way to call back into UART and explicitly send an interrupt, because I'll be running a separate terminal thread for proper event-driven approach. This asks for a bit of redesign to chardev_t api

Okay, I think I got a suitable api, inspired from `hid_dev.h` and `tap_linux.c`:
```
#ifndef CHARDEV_H
#define CHARDEV_H

#include "rvvmlib.h"

typedef struct chardev chardev_t;

typedef struct {
    void on_input_available(chardev_t* dev);
    void on_output_available(chardev_t* dev);
    void* userdata;
} chardev_watcher_t;

struct chardev {
    size_t (*read)(void* buf, size_t nbytes, void* userdata);
    size_t (*write)(const void* buf, size_t nbytes, void* userdata);
    void (*destroy)(void* userdata);
    chardev_watcher_t* watcher;
    void* userdata;
};

static inline size_t chardev_read(chardev_t* dev, void* buf, size_t nbytes)
{
    return dev->read(buf, nbytes, dev->userdata);
}
static inline size_t chardev_write(chardev_t* dev, const void* buf, size_t nbytes)
{
    return dev->write(buf, nbytes, dev->userdata);
}
static inline void chardev_destroy(chardev_t* dev)
{
    return dev->destroy(dev->userdata);
}
static inline void chardev_input_available(chardev_t* dev)
{
    return dev->watcher->on_input_available(dev);
}
static inline void chardev_output_available(chardev_t* dev)
{
    return dev->watcher->on_output_available(dev);
}

#endif
```

A `term` or `tcp` chardev can create a thread to do select/poll loop, notify the watcher (ns16550a or chardev_mux) by `chardev_input_available` and `chardev_output_available`.

One problem I now have is to shutdown/thread_join the loop of `chardev_term`, does a `pipe` (I see in `tap_linux.c` , `poll` is shutdown by a `pipe`) suitable to shutdown the POSIX `select` loop?  How about Windows?

For `chardev_tcp`, I guess `net_poll_close` is enough.

WDYT, thank you!

LekKit:
> One problem I now have is to shutdown/thread_join the loop of `chardev_term`, does a `pipe` (I see in `tap_linux.c` , `poll` is shutdown by a `pipe`) suitable to shutdown the POSIX `select` loop? How about Windows?

For `chardev_tcp`, look at `tap_user` and how it uses `net_tcp_sockpair()`. It creates a local socket pair - basically a portable `pipe()` on anything that has sockets.

For `chardev_term`, the terminal thread killing implementation is likely to be very obscure, as all things under Windows. I am fine if you don't want to handle it properly right now and you may continue using a sleep-poll loop under Windows till I come with a better solution. You may also add a `update()` call sorta to the chardev interface because it's possible there is no other way on Windows...

LekKit:
> For `chardev_tcp`, I guess `net_poll_close` is enough.

Maybe, but be cautious (This wasn't tested well). Thing is, `net_poll_close()` frees a `net_poll_t*` pointer and if you do anything to it from another thread afterwards you have a UAF and other nasty consequences. 

As I already said a sockpair is a much cleaner and better approach than RCU madness and guessing if each of `net_poll` implementations acts well upon closing with waiters on it.

ghost:
Hello, thank you for the tips.  Now it can shutdown properly, yet there are still issues I know:
- The term and tcp thread consume 100% cpu, it seems when polling the output is always available, or if input available but not consumed, then the loop will call callbacks to waste cpu..  Should we add a sleep here, at what interval?
- When a tcp client is disconnected, its sock won't pop from net_poll_wait immediately, only when try to write to it.  I think this is not a big deal..

Also I'm bad at naming things, feel free to suggest alternative names or api structures, or make changes by your pace.

LekKit:
> * The term and tcp thread consume 100% cpu, it seems when polling the output is always available, or if input available but not consumed, then the loop will call callbacks to waste cpu..  Should we add a sleep here, at what interval?

There needs to be more logic concerning "input available but the guest haven't looked yet" and "output is available but we do not care"
Generally you should poll for input, and then remove it from watch set until the guest empties the input buffer and `read(stdin)` returns 0. So basically edge-triggered input availability detection sorta.
Polling for output availability should only happen if the guest enabled the "TX" interrupt, and also should be edge-triggered (Reset only after a write).

> * When a tcp client is disconnected, its sock won't pop from net_poll_wait immediately, only when try to write to it.  I think this is not a big deal..

What do you mean? You want to know how to detect a disconnect?
Disconnected clients always post an event, and when you `recv()` or `send()` on them, you get 0 in return. It is a sign of a socket disconnect. Example:

```cpp
ev_count = net_poll_wait(poll, events, NET_POLL_INF);
for (size_t i=0; i<ev_count; ++i) {
    net_sock_t* sock = // some way to get socket from events[i].data context
    size_t received = net_tcp_recv(sock, ...);
    if (!received) {
        // this client has disconnected!
    }
}
```

LekKit:
> Also I'm bad at naming things, feel free to suggest alternative names or api structures, or make changes by your pace.

I am just a bit overwhelmed by the things going on :D, and I am always asking you to change some stuff.. Because I am afraid of possible merge conflicts or sth, or just don't like disturbing people in general.

I need to do 1 simple change to `ns16550a.c` upstream, actually include `<sys/select.h>` because on Serenity it is not automatically included elsewhere and it raises a warning. May I do this? You'll need to rebase

Another thing that I'd like to be changed in current chardev design. It is manually freed, but all the current RVVM devices have a notion of `remove()` callback and are automatically freed when the owning VM is destroyed. This significantly simplifies memory management for library user, and keeps RVVM leak-free fool-proof.
I'd like that `ns16550a_remove()` calls `chardev_free()` so they are freed together.

ghost:
> Generally you should poll for input, and then remove it from watch set until the guest empties the input buffer and read(stdin) returns 0. So basically edge-triggered input availability detection sorta.

Cool, so this is what edge-triggered mean, I'll try it.

> You'll need to rebase
No problem for me :)

> I'd like that ns16550a_remove() calls chardev_free() so they are freed together.
Got it.

I'll keep going according the advices and back here later, thank you!

ghost:
@LekKit Hello, I have get the terminal backend working with a event-driven interface.

- edge-triggered is emulated by manipulate fd_sets via a pipe `ctl` (or atomic bits on windows).
- on windows, the terminal_thread is a loop with `sleep_ms(1)`, since I find no way to select for stdio..
- output callback is enabled only when the `THR` inturrupt is needed, and it can be changed between enabled and disabled.  So `chardev_watch` can be called multiple times.
- no buffer implemented now, so `select` may wake up for each char read/write, I think this can be improved later..

I'd postpone the code for chardev_tcp now, maybe for the android terminal, a `posix_openpt` implemented pseudo-terminal device is better?  @fish4terrisa-MSDSM

fish4terrisa-MSDSM:
> @LekKit Hello, I have get the terminal backend working with a event-driven interface.
> 
> * edge-triggered is emulated by manipulate fd_sets via a pipe `ctl` (or atomic bits on windows).
> * on windows, the terminal_thread is a loop with `sleep_ms(1)`, since I find no way to select for stdio..
> * output callback is enabled only when the `THR` inturrupt is needed, and it can be changed between enabled and disabled.  So `chardev_watch` can be called multiple times.
> * no buffer implemented now, so `select` may wake up for each char read/write, I think this can be improved later..
> 
> I'd postpone the code for chardev_tcp now, maybe for the android terminal, a `posix_openpt` implemented pseudo-terminal device is better? @fish4terrisa-MSDSM

Hmm,a pty device is better so that I dont need to run the `resize` command(included in the xterm package) to correct the size of the terminal.

fish4terrisa-MSDSM:
> @LekKit Hello, I have get the terminal backend working with a event-driven interface.
> 
> * edge-triggered is emulated by manipulate fd_sets via a pipe `ctl` (or atomic bits on windows).
> * on windows, the terminal_thread is a loop with `sleep_ms(1)`, since I find no way to select for stdio..
> * output callback is enabled only when the `THR` inturrupt is needed, and it can be changed between enabled and disabled.  So `chardev_watch` can be called multiple times.
> * no buffer implemented now, so `select` may wake up for each char read/write, I think this can be improved later..
> 
> I'd postpone the code for chardev_tcp now, maybe for the android terminal, a `posix_openpt` implemented pseudo-terminal device is better? @fish4terrisa-MSDSM

I have just tried the patch and it didn\`t work .The size of the tty is still wrong...a `posix_openpt` implemented pty might work better,but I\`m not so sure.(Sorry for my poor English)

fish4terrisa-MSDSM:
The terminal just looks like this,and i have to run `resize` to fix the size.
Screenshot_20230428_200216

fish4terrisa-MSDSM:
And connect the terminal on tcp might prevent the user from start multiple VMs, or they have to choose port. Maybe unix socket might be better for unix-like ports? (The path to the unix socket file tells the usage of it, but a tcp port can\`t)

ghost:
@fish4terrisa-MSDSM 
> I have just tried the patch and it didn`t work .The size of the tty is still wrong...

Well, it seems the serial term have to be resized manual, see:
https://wiki.archlinux.org/title/Working_with_the_serial_console#Resizing_a_terminal
https://unix.stackexchange.com/questions/106644/how-to-change-the-width-of-remote-serial-console

> a posix_openpt implemented pty might work better,but I`m not so sure.

Okay, now I see your code use pty with rvvm stdio in `https://github.com/fish4terrisa-MSDSM/archriscv-term/blob/flyingfish/archriscv-app/app/src/main/jni/terminal_jni.c`.
I think a `posix_openpt` backend for rvvm serial can add support of multiple ttys to it.


> And connect the terminal on tcp might prevent the user from start multiple VMs, or they have to choose port. Maybe unix socket might be better for unix-like ports

Yes, use a tcp socket will need more works for ports..  and a unix socket backend should work similiar to pty.

LekKit:
> Hmm,a pty device is better so that I dont need to run the `resize` command(included in the xterm package) to correct the size of the terminal.

The UART tty is not auto-resizable at all because the guest (Linux) has no way to know the remote console was resized (An X terminal emulator would send a `SIGWINCH` signal, but no such signal facility over the UART.

You may add `resize` as a boot service, but repeated changes to the TTY size would require polling it from the guest.

It is theoretically possible to send some VT escapes into the guest to signal the new console size, but most likely will require some guest interoperation still. SSH is our best bet(?)

fish4terrisa-MSDSM:
> > Hmm,a pty device is better so that I dont need to run the `resize` command(included in the xterm package) to correct the size of the terminal.
> 
> The UART tty is not auto-resizable at all because the guest (Linux) has no way to know the remote console was resized (An X terminal emulator would send a `SIGWINCH` signal, but no such signal facility over the UART.

Ok, i will try to find a way to resize the terminal when boot the system.

fish4terrisa-MSDSM:
> 
> Okay, now I see your code use pty with rvvm stdio in `https://github.com/fish4terrisa-MSDSM/archriscv-term/blob/flyingfish/archriscv-app/app/src/main/jni/terminal_jni.c`. I think a `posix_openpt` backend for rvvm serial can add support of multiple ttys to it.
> 
I just borrowed these codes from alpine-term,and i dont\`t know much about it .The only things i know is that it gets the stdio of the program and link it to a terminal.I will work out what these codes have done later.However , I think opening new terminals and run `socat` inside it to connect to sockets which binded to serial ttys  is a easier choice.(Thats the method I used to support multiple ttys for QEMU)

fish4terrisa-MSDSM:
> > Hmm,a pty device is better so that I dont need to run the `resize` command(included in the xterm package) to correct the size of the terminal.
> 
> The UART tty is not auto-resizable at all because the guest (Linux) has no way to know the remote console was resized (An X terminal emulator would send a `SIGWINCH` signal, but no such signal facility over the UART.
> 
> You may add `resize` as a boot service, but repeated changes to the TTY size would require polling it from the guest.
> 
I have tried adding it as a systemd service ,but it don\`t work .Adding it to the bashrc or inside `/etc/profile.d/`might be a good idea(?)
> It is theoretically possible to send some VT escapes into the guest to signal the new console size, but most likely will require some guest interoperation still. SSH is our best bet(?)
SSH might work , but that method have some performance cost (I have tried it with QEMU, it\`s a bit slow and randomly the connection breaks down)

LekKit:
The current implementation is good enough to be in `staging`, but afterwards I will do some tiny changes (rewrite win32 console and maybe slight refactoring). Afterwards we will decide what to do with PTY/TCP chardevs, they will be a separate PR if anything

Are you OK with merging it for now? The discussion here may move to actual discussions or to issue #86

ghost:
> Are you OK with merging it for now? The discussion here may move to actual discussions or to issue #86

Yes, thank you!  And next I'll work on a libretro port and tcp/pty/libvterm chardev backends.

LekKit:
FYI about win32:
- `putc()` is very slow on Windows, but `WriteConsoleW()` is faster (However the Windows console is still very slow in general). It needs UTF16 strings, but UART will feed individual bytes so I have to buffer them and do streaming UTF8 conversion.
- It might be wise to buffer all output bytes and output them via the chardev IO thread instead. This way we eliminate the individual per-character syscalls and prevent the guest CPU stalls (This is currently the case on v0.5, and this effect is enough to slow down guest Linux boot by ~5 seconds on Windows host)

LekKit:
Okay, some more news:
- Changing `chardev_term` read/write to operate on a userspace FIFO ring instead of syscalls directly, and doing IO in whatever separate thread gives HUGE boost to uart speed (But requires a lot of rewrite)
- Went too far with the refactoring and we now may reuse `update` callbacks in chardev API if we don't wish to use a separate thread (It seems that a dedicated uart IO thread is a bit of an overkill idea from me for a lot of usecases (win32), sorry!)

Will be pushing this soon but if anyone doesn't like what I did, I will gladly listen to figure the best way collaborately

PR #95, opened 2023-04-29
chardev_term: Add POSIX fd backend

ghost:
I think this can help with #86 

- First run socat to setup some pty/pts:
  `socat pty,rawer,link=/tmp/pts1 stdio,rawer`
  `socat pty,rawer,link=/tmp/pts2 stdio,rawer`

- Then in `main.c`, create fd chardev with those pts files:
```
modified   src/main.c
@@ -21,7 +21,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 #include "rvvmlib.h"
 #include "utils.h"
-
+#include "fcntl.h"
 #include "devices/clint.h"
 #include "devices/plic.h"
 #include "devices/ns16550a.h"
@@ -213,7 +213,12 @@ static int rvvm_main(int argc, const char** argv)
     pci_bus_init_auto(machine);
     i2c_oc_init_auto(machine);
 
+    int pts1 = open("/tmp/pts1", O_RDWR);
+    int pts2 = open("/tmp/pts2", O_RDWR);
+    ns16550a_init_auto(machine, chardev_term_fd_create(pts1, pts1));
+    ns16550a_init_auto(machine, chardev_term_fd_create(pts2, pts2));
     ns16550a_init_term_auto(machine);
+
     rtc_goldfish_init_auto(machine);
     syscon_init_auto(machine);
```

- Finally, run rvvm, maybe: `./rvvm fw_jump.bin -k linux -cmdline 'console=ttyS1'`
It seems opensbi always use ttyS0 though.

### DIFF
--- a/src/devices/chardev.h
+++ b/src/devices/chardev.h
@@ -81,6 +81,7 @@ static inline void chardev_notify(chardev_t* dev, uint32_t flags)
 
 // Built-in chardev implementations
 
-PUBLIC chardev_t* chardev_term_create(void);
+PUBLIC chardev_t* chardev_term_create(void); // stdio
+PUBLIC chardev_t* chardev_term_fd_create(int rfd, int wfd); // POSIX fd
 
 #endif
--- a/src/devices/chardev_term.c
+++ b/src/devices/chardev_term.c
@@ -78,6 +78,7 @@ typedef struct {
     uint32_t tx_size;
     char rx_buf[256];
     char tx_buf[256];
+    int rfd, wfd;
 } chardev_term_t;
 
 static uint32_t term_update_flags(chardev_term_t* term)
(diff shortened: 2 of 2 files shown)


LekKit:
I have some none-upstreamed changes upon `chardev_term` which reuse ringbuf properly and these will lead to merge conflicts (~~I can help to resolve~~ I will merge it as is and rebase on my side).

Perhaps makes sense to also wrap PTY opening into `chardev_term` api itself?
Sth like this: `chardev_pty_create("/tmp/ptsX")`

LekKit:
>   It seems opensbi always use ttyS0 though.

It might be using FDT `chosen-stdout` property which is tied to first UART in the current design (And it is troublesome to overcome).
In general it might not matter for now unless you want to debug things, and in any case you may omit initializing the first UART on the terminal.

ghost:
> I will merge it as is and rebase on my side

Okay, thank you!

> Perhaps makes sense to also wrap PTY opening into chardev_term api itself? Sth like this: chardev_pty_create("/tmp/ptsX")

Does this mean open the master pty with `posix_openpt`, `unlockpt`, and `grantpt`, and then symlink its ptsname to a file (/dev/ptsX to /tmp/ptsX)?
I think both should works for @fish4terrisa-MSDSM 's usecass.

LekKit:
> Does this mean open the master pty with `posix_openpt`, `unlockpt`, and `grantpt`, and then symlink its ptsname to a file (/dev/ptsX to /tmp/ptsX)? I think both should works for @fish4terrisa-MSDSM 's usecass.

Yes, as this will expose this functionality as a `librvvm` feature rather than hacking around FDs in `main.c` which is more convenient in the long term, omits touching system APIs in unwanted places (And may be used from JNI for that usecase)

PR #96, opened 2023-04-30
Add libretro core

ghost:
- Build with `make CC=gcc BUILD_LIBRETRO=1`
- Run with `retroarch -L release.linux.x86_64/rvvm_libretro.so foo.rvvm`

foo.rvvm (within the same directory of firmware/image files) can be:
```
bootrom=fw_jump.bin
kernel=linux_6.1
nvme1=rootfs.ext2
```
Tested with the doom image, serial terminal not implemented now..


LekKit:
- Is it possible for `libretro` to be part of `librvvm` shared lib? Or are there implications I am unaware of?
- The Makefile changes are a tiny bit messy, there shouldn't be a branch around the `all` target but rather a lists of targets as a variable. Or better, a `libretro` target so we can `make all lib libretro` and have the whole squad of them made
- Perhaps use `src/framebuffer.h` API instead of `fb_window.h` which is kinda middle ground between public API and internall RVVM mess for window systems
- Is it portable to Windows/Non-Linux POSIXes? Just a quick question whether it needs testing maybe
- What about `src/bindings/libretro/` path? I have a yet not upstreamed `src/bindings/jni/` so maybe they can coexist there
- Bump the version to `0.6-git`, the `v0.5` times are off and it was released as stable (Mainly to give RVVM at least *some* stable release which is known to not regress or break API every day :D)

Some further things revealed (Retroarch related):
- Switching focus from Retroarch shows it's been paused, but apparently `rvvm_pause_machine()` is not called so it keeps running
- A lot of keypresses are not passed to RVVM at all(?) but instead are handled by Retroarch (Switching fullscreen, etc), typing is barely possible. Some option changes that maybe?
- Pressing ESC twice quits Retroarch, which may or may not be nice for some RVVM uses (If it can't be worked around don't worry)

If there are any other features in Retroarch you might want to implement, ask here)
I see there is fast-forward functionality which may be implemented by scaling the `rvtimer` frequency (But not actually exposed in the RVVM API)

ghost:
> Is it possible for libretro to be part of librvvm shared lib? Or are there implications I am unaware of?

Well, I guess it can, but a libretro core (shared library, or static library for some specified platforms like 3ds/psp) should only expose 'retro_' APIs and not linked with externel dependencies like libX11/SDL.

> The Makefile changes are a tiny bit messy, there shouldn't be a branch around the all target but rather a lists of targets as a variable.

To avoid link with libX11/SDL for libretro, `USE_FB` need to be set as 0, but a normal target should want it.  The
OBJS are shared between BINARY and LIBRETRO target, so in a run of 'make', I think you can only choose to build for libretro or for other, not both...

> Perhaps use src/framebuffer.h API instead of fb_window.h

Okay, make sense, I'll look into it, thanks!

> Is it portable to Windows/Non-Linux POSIXes

Yes it should be, retroarch support many kindof platforms (android, psp, ios, 3ds, etc),  but I only tested it on Linux now.  Later I'll upstream this (by using their CI to build and distribute the core) to libretro-super and hope if there are more interested users can help with portable tests..

> What about src/bindings/libretro/ path?

Well, I think the libretro target is not a binding, it's like `rvvm` as an entrypoint, not for future development.

> Bump the version to 0.6-git

Okay.

> Switching focus from Retroarch shows it's been paused, but apparently rvvm_pause_machine() is not called so it keeps running

I don't know if there is a hook point to call `rvvm_pasue_machine`, when retroarch is paused, it just stop to call `retro_run`..


> A lot of keypresses are not passed to RVVM at all.

There is "Game focus" mode to pass all keyboard events (except a key to quit the mode) to the core, mine's game focus hotkey is binded to "Scroll Lock".

> Pressing ESC twice quits Retroarch

Yes, in game focus mode it just pass ESC to guest.

 > If there are any other features in Retroarch you might want to implement, ask here.  I see there is fast-forward functionality...

Sure, I think the next should be audio support.  
And retroarch also have netplay and rewind feature, rewind run the world backwards, I imaged that's useful for debug/fun...
To implement netplay/rewind all state (at least memory and cpu registers) needed to serializable (retro_serialize/retro_unserialize).

X547:
> Is it possible for `libretro` to be part of `librvvm` shared lib? Or are there implications I am unaware of?

I would prefer to have `librvvm` without `libretro` parts. It is preferable to avoid extra dependencies in UserlandVM and similar projects.

LekKit:
> Well, I guess it can, but a libretro core (shared library, or static library for some specified platforms like 3ds/psp) should only expose 'retro_' APIs and not linked with externel dependencies like libX11/SDL.

Ah, makes sense then. I think we can split it by copying `OBJS` into separate variable and filtering gui sources from there. I will try to do it and give you a patch if it works.

> Well, I think the libretro target is not a binding, it's like `rvvm` as an entrypoint, not for future development.

It makes some sense since it's technically a wrapper around RVVM API, hence if we change something in RVVM API, it is a good idea to also walk over bindings and test them. It also would simply look like a nicer directory structure.
 
> I don't know if there is a hook point to call `rvvm_pasue_machine`, when retroarch is paused, it just stop to call `retro_run`..

Hm. Maybe we can record a timeout somewhere to shutdown the VM thread, but of course handling a  pause event would be much cleaner. Pity if Retroarch doesn't support that.

> To implement netplay/rewind all state (at least memory and cpu registers) needed to serializable (retro_serialize/retro_unserialize).

Can it use some special structure beneath it (or file backed for that matter)?
RVVM state serializing is very WIP and it also consumes significant amount of RAM/disk space. I am not sure how netplay is related to it, but if it means sending VM state over network that would work horrible here.

LekKit:
> I would prefer to have `librvvm` without `libretro` parts. It is preferable to avoid extra dependencies in UserlandVM and similar projects.

You are right, but the idea wasn't to enable it by default. It would just simplify things a bit if `libretro` port would be a superset of `librvvm` so you don't copy same code in two packages and one may replace the other (For example in AUR).

AFAIK it does not bring any new dependencies or noticeable code bloat for that matter, just a few new symbols
```
$ ldd rvvm_libretro.so 
        linux-vdso.so.1 (0x00007ffcc01f8000)
        libatomic.so.1 => /usr/lib/libatomic.so.1 (0x00007fe1ce68f000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007fe1ce5a7000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007fe1ce3c0000)
        /usr/lib64/ld-linux-x86-64.so.2 (0x00007fe1ce719000)
$ du -h rvvm_libretro.so 
320K    rvvm_libretro.so
```

Altho, I want to bring JNI bindings into `librvvm`. They do not bring any observable code bloat (a few tiny functions exposed to bridge it) and `jni.h` would be self-contained in the repo so you likely won't even notice it (Or we can disable `USE_JNI` on Haiku target if you wish).

Ultimately I'm with you on "no dependencies" goal, and beyond the optional stuff like SDL it is not going to change. I do not include `libvterm` support for the same reason (But the inhouse VT will likely be there).
Simply exposing additional features or symbols I believe does not add much to the complexity of it

ghost:
> Simply exposing additional features or symbols I believe does not add much to the complexity of it

Well, i see that most libretro cores use a linker script to hide all symbols expect 'retro_*' ones,  so I do the same.
I think that's required to avoid symbols clash with retroarch or about static link, not sure though...
If that's the case, the libretro core can't not used as a shared library as librvvm.

ghost:
There is a segfault issue which would happend after "Close Content" in RetroArch.
I think it's due to threads running `threadpool_worker` are still alive after `rvvm_free_machine`.

LekKit:
> Well, i see that most libretro cores use a linker script to hide all symbols expect 'retro_*' ones, so I do the same. I think that's required to avoid symbols clash with retroarch or about static link, not sure though... If that's the case, the libretro core can't not used as a shared library as librvvm.

I see. Makes sense, altho not really nice (There is no way `librvvm` clashes with anything).
RVVM is also using `-fvisibility=hidden` on purpose which hides everything not marked with `PUBLIC` attribute. It is perhaps possible to reuse it here.

> There is a segfault issue which would happend after "Close Content" in RetroArch. I think it's due to threads running `threadpool_worker` are still alive after `rvvm_free_machine`.

~~Hm, could not reproduce.~~ Okay, I see.
There is some time the threadpool workers are still alive even tho there are no tasks. This is to have them warmed up and use a single `pthread_cond` wakeup to kick them.
They are killed on `atexit()` which is properly ran when `rvvm-cli` exits, but not when we unload `librvvm`.
This asks for some `rvvm_deinit()` call and internal mechanism mirroring the use of `atexit()`

P.S. Possible way to handle pause is `retro_throttle_state`?

ghost:
Um, after looking into libretro's doc [1] and CI infra, I think use CMake is easier.

- A specified `.gitlab-ci.yml` file is needed for its CI (eg: [2]), libretro will mirror this repo in its GitLab (eg: [3]), which has access to its ci-templates [4] and runner.
- The `.gitlab-ci.yml` can choose between make or cmake templates, and use `variables` to customize some paths or build options.
  but the make templates assume the built core is in the same directory with Makefile, and the excepcted Makefile is a bit messy..
  https://git.libretro.com/libretro-infrastructure/ci-templates/-/blob/master/linux-x64.yml#L25
  https://github.com/nbarkhina/numero/blob/master/Makefile.libretro

- the cmake templates look more clean.
https://git.libretro.com/libretro/wasm4/-/blob/main/.gitlab-ci.yml
https://git.libretro.com/libretro/wasm4/-/blob/main/runtimes/native/CMakeLists.txt

Should I drop the `Makefile` changes in this PR, and modify `CMakeLists.txt` instead?

[1] https://docs.libretro.com/development/cores/developing-cores/#add-your-core-to-libretro-infrastructure
[2] https://github.com/nbarkhina/numero/blob/master/.gitlab-ci.yml
[3] https://git.libretro.com/libretro/numero
[4] https://git.libretro.com/libretro-infrastructure/ci-templates

LekKit:
> Should I drop the `Makefile` changes in this PR, and modify `CMakeLists.txt` instead?

If that means `rvvm_libretro` core could be built in `libretro` CI, it might simplify life a lot so go for it. I will invent something for Makefile later on (Already have some ideas).

src/libretro/libretro.c
```diff
@@ -0,0 +1,534 @@
...
+static retro_input_state_t input_state_cb;
+static retro_environment_t environ_cb;
+static rvvm_machine_t *machine;
+static thread_ctx_t *vm_thread;
+static fb_ctx_t vm_fb;
+static hid_keyboard_t *vm_keyboard;
+static hid_mouse_t *vm_mouse;
+static struct {
+    int smp;
+    int mem;
```
LekKit: This causes UB and unexpected issues when setting RAM to >=2048M (It overflows on shift later on). Use `size_t` here (Same goes to cores?).

Rule of thumb: never use signed types unless it is mandatory to handle negative values. For anything that counts objects or size the go-to type is `size_t`.

It might be wise to handle the size suffixes in parser

src/libretro/libretro.c
```diff
@@ -0,0 +1,534 @@
...
+        if (v == NULL) {
+            log_cb(RETRO_LOG_ERROR, "Invalid option: %s\n", line);
+            continue;
+        }
+        if (strcmp(k, "mem") == 0) {
+            machine_opts.mem = atoi(v);
+            continue;
+        }
+        if (strcmp(k, "smp") == 0) {
+            machine_opts.mem = atoi(v);
```
LekKit: Copy-paste error

src/libretro/libretro.c
ghost: Done, thank you.

src/libretro/libretro.c
ghost: Done, thank you!

ghost:
Changed to use `CMakeLists.txt` and add a `.gitlab-ci.yml`.
After merged I'll ask libretro upstream how to integrate this, thank you!

LekKit:
Is there an easy way to run this GitLab CI now and see what targets build and what don't? I mean, without disturbing libretro team for now and to just test things out.

ghost:
> Is there an easy way to run this GitLab CI now and see what targets build and what don't?

I don't know, asking now by opening a PR in libretro-super..

LekKit:
Question: What about dropping `nvme0`, `nvme1` and instead instantiating devices on each of duplicated `nvme` config lines?
You seem to have followed CLI way of parsing arguments but the CLI implementation is already a bit cursed.

I am asking because maybe later on the config syntax will be adopted by RVVM in general and not simply for `libretro` port, and it would make sense to set on some good config conventions.
An ultimate end goal would be to have some standardized way of parsing config/arguments which shares most of the code and conventions, so that a single source of documentation covers both

LekKit:
Some more things to note:
- U-Boot and the Haiku bootloader expect the UART to be present, and it it is not, they stop working. It is therefore mandatory to attach it, and it is even defined so in RISC-V Platform Spec. If you don't want output, it is legal to pass `NULL` chardev and the uart output will be silenced (Cool huh). Later on there could be some chardev for `libretro` but for now this works.
- You seem to not handle firmware/kernel loading errors or NVMe attach errors, if passed an illegal path in config it just ignores that and leaves no notice that sth is wrong to `libretro`
- Perhaps put the RTL8169 in there under `#ifdef USE_NET` otherwise there is no networking

P.S. Never knew I needed these so much till today...
<details>
  <summary> Cool aesthetic screenshots </summary>
  
image

image

image

image

</details>

ghost:
> What about dropping nvme0, nvme1 and instead instantiating devices on each of duplicated nvme config lines?

Okay, done.  For now I set a limit to allow max 4 nvme devices.

> An ultimate end goal would be to have some standardized way of parsing config/arguments which shares most of the code and conventions, 

Yes, that will be great.  I'm not good at write config parser..

> U-Boot and the Haiku bootloader expect the UART to be present.

Okay, I set a `NULL` chardev now.  Maybe they can work with framebuffer as chosen output device (I tried to put the framebuffer node under chosen, but it does not work..)?

> Perhaps put the RTL8169 in there under #ifdef USE_NET otherwise there is no networking

Done, thanks.

LekKit:
I kinda still want it to be in `src/bindings/libretro` because of the third party header, that way I can disable static analyzers/linters on `src/bindings` entirely.
The directory structure would also make a bit more sense since `libretro` is not exactly a part of RVVM but a "binding" atop it, as opposed to `src/rvjit`, `src/devices`. I very much hope we can agree on this (Same as there is no reason to throw `src/jni` in there like that).
Afterwards this is good to merge.

ghost:
> I kinda still want it to be in src/bindings/libretro because of the third party header, that way I can disable static analyzers/linters on src/bindings entirely.

No problem!

ghost:
Next we may add some document about RVVM in https://github.com/libretro/docs#adding-a-new-core

PR #105, opened 2023-09-01
nvme_io_cmd refactor and optimization

ghost:
Eliminated redundant variable declarations and assignments. Simplified error handling by directly checking the return values of blk_write and blk_read. Combined some conditionals to reduce code duplication, removed unnecessary comments and improved code readability.


g00g1:
@grasek, with all due respect for your eagerness to help this project I should warn you that your current code just not works and fails all build test. Please check your code locally before doing such pull requests as you distract the developer from important staff.

LekKit:
This change makes the code unreadable, and erroneously submits 2 NVMe command completion entries upon an error

PR #106, opened 2023-09-10
Update rvvm.c

ghost:


g00g1:
Dear @grasek, as per your previous pull request, the code you have contributed is not working at all. Please **do not** distract the developer from important tasks by bombing their repository with meaningless changes.

PR #115, opened 2023-10-27
Update README.md

JanumalaAkhilendra:
Fixed Some Grammatical errors in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ RISC-V CPU & System software implementation written in С
 ## What's working
 [![Demo](https://img.shields.io/badge/Check%20it%20out-WASM%20Demo-red?style=for-the-badge)](https://lekkit.github.io/test/index.html)
 - Passes RISC-V compliance/torture tests for both RV64 & RV32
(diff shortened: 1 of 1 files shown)


000exploit:
I consider only the "trademark" as a good change. Fixing "Poweroff" to "Power off" and "virtio" to "virtual" looks like nonsense. In the first case, a specific device is meant, and in the second case, a certain class of virtual devices is meant, called "virtio" (e.g. virtio-net, virtio-blk, etc).

Mr0maks:
> I consider only the "trademark" as a good change. Fixing "Poweroff" to "Power off" and "virtio" to "virtual" looks like nonsense. In the first case, a specific device is meant, and in the second case, a certain class of virtual devices is meant, called "virtio" (e.g. virtio-net, virtio-blk, etc).

Looks like a bot. Also trademark writed because in Russian its a two separated words.

LekKit:
Choosing "trademark" over "trade mark" is just a difference of spelling in UK and US, same as color/colour.

PR #129, opened 2024-03-15
Drop "OS" from Haiku's name.

OscarL:
Nitpicking, but... same as Linux, FreeBSD, Serenity, etc... it is just Haiku.

(when listing it among other OSes' names, no need for the qualifier).


LekKit:
I am actually unsure whether it's a on official naming scheme or not. Upstream website domain is haiku-os.org, which also says "Haiku OS" in the title; Same story with Reddit community dedicated to the project. Many articles mention "Haiku OS" specifically, but the wikipedia page name is simply "Haiku".

Perhaps @X547  can enlighten us about the recommended naming scheme?

X547:
Official name is "Haiku" (without "OS").

https://www.haiku-os.org/about/faq#why-isnt-it-called-haikuos

LekKit (approved):

PR #171, opened 2025-02-03
Implement wayland backend

0xCatPKG:
Supported features:
- Rendering;
- Keyboard input;
- Mouse input;
- Server-side decorations (Compositor-dependent, requires xdg-decoration protocol);
- Mouse input grab (Compositor-dependent, requires pointer-constraints and relative-pointer protocols, supported by most compositors)
-  Keyboard input grab (Compositor-dependent, supported by major compositors, like Mutter, KWin, Sway, Cosmic and Hyprland)
- Window title changes
- Full fullscreen support (can be activated from vm or if size of window >= output size);
- Closing of windows

Not supported yet:
- Global removal (outputs and seats not properly freed if compositors removes them

Requirements:
libwayland-client.so (Dynamic linking)
libxkbcommon.so (Dynamic linking)
Compositor with xkb keymaps support (Currently all compositors use xkb, so no problems here)


X547:
> Server-side decorations (Compositor-dependent, requires xdg-decoration protocol);

Is it possible to also implement https://wayland.app/protocols/kde-server-decoration? There is a controversy with xdg-decoration so it is not implemented in GTK.

0xCatPKG:
image
image

@X547 can you list me potential users of that protocol?

X547:
> @X547 can you list me potential users of that protocol?

My Wayland server implementation for Haiku: https://github.com/X547/wayland-server . Or maybe I should implement xdg-decoration instead, but I do not know any important software that implement it, but do not implement KDE decoration protocol.

Very low priority problem anyway.

0xCatPKG:
kde-server-decorations protocol is deprecated.

Currently, no one who implements SSD via protocols only supports kde_server_decorations and does not support xdg_decoration. In any case, xdg_decoration **doesn't** guarantee that the server supports SSD, it's just a way of communicating between client and compositor what everyone wants and can do.

PR #176, opened 2025-02-12
Optimize orc.b emulation

purplesyringa:
This patch optimizes the platform-independent orc.b computation. Here's how this works:

Consider any byte `0bABCDEFGH` of the input. In `(0bABCDEFGH | 0x80) - 1`, the top bit stays `1` if any of the bits BCDEFGH are non-null, and resets to `0` otherwise. ORing the result with the byte `0bABCDEFGH` ensures that the top bit stays `1` whenever *any* of the bits is non-empty. We then mask the bit and broadcast it to the whole byte by multiplying it with `0xFF` (or, rather, simulating that with shifts).


LekKit (approved):
Much appreciated <3

PR #183, opened 2025-05-12
allow wasm terminal input

saeziae:
allow terminal input when used with xterm.js

yes it just works.

imatge

### DIFF
--- a/src/devices/chardev_term.c
+++ b/src/devices/chardev_term.c
@@ -11,7 +11,7 @@ file, You can obtain one at https://mozilla.org/MPL/2.0/.
 #include "chardev.h"
 #include <stdio.h> // For fputs(), setvbuf()
 
-#if (defined(__unix__) || defined(__APPLE__) || defined(__HAIKU__)) && !defined(__EMSCRIPTEN__)
+#if (defined(__unix__) || defined(__APPLE__) || defined(__HAIKU__)) || defined(__EMSCRIPTEN__)
 #include <fcntl.h>    // For open()
 #include <sys/time.h> // For select()
 #include <termios.h>  // For tcgetattr(), tcsetattr(), etc
--- a/src/devices/chardev_term.c
+++ b/src/devices/chardev_term.c
@@ -11,7 +11,7 @@ file, You can obtain one at https://mozilla.org/MPL/2.0/.
 #include "chardev.h"
 #include <stdio.h> // For fputs(), setvbuf()
 
-#if (defined(__unix__) || defined(__APPLE__) || defined(__HAIKU__)) || defined(__EMSCRIPTEN__)
+#if defined(__unix__) || defined(__APPLE__) || defined(__HAIKU__)
 #include <fcntl.h>    // For open()
 #include <sys/time.h> // For select()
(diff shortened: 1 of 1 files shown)


LekKit:
Btw the `|| defined(__EMSCRIPTEN__)` is not even needed then, as it implies `__unix__`.

How do I use xterm.js itself on the RVVM demo page at https://lekkit.github.io/test/index.html though? X)

saeziae:
> Btw the `|| defined(__EMSCRIPTEN__)` is not even needed then, as it implies `__unix__`.

verified, modified

> How do I use xterm.js itself on the RVVM demo page at https://lekkit.github.io/test/index.html though? X)

i was exploring this issue just because i was cooking it :3

welcome to see <https://vm.estela.moe>

project at <https://github.com/saeziae/homepage-rvvm-wasm>, it's chaotic tho.

LekKit:
That's pretty!

LekKit:
Basically the reason that I originally disabled input on Emscripten, was that it apparently reported `stdin` as ready in `select()`, but at the same time `read()` blocked and the event thread locked up, if it can be solved and an actual terminal implemented, then it's the best thing ever

saeziae:
by the way i don't know about the template (shell.html), so in that project i modified Makefile too, to get mjs file as output for doing manual hook

LekKit:
> i modified Makefile too, to get mjs file as output for doing manual hook

That's probably what I'm gonna do too with a coming batch of Makefile changes, I guess there's no benefit from the Emscripten-provided shell.html as it's fairly limited anyways.

LekKit:
See updated https://lekkit.github.io/test/

Note that I had to patch `xterm-pty`, specifically the `emscripten-pty.js` library that is linked into the final `rvvm_wasm32.mjs` module. It now somewhat works without neither `ASYNCIFY` nor `PROXY_TO_PTHREAD` Emscripten option, because Asyncify slows things down a great deal, and PROXY_TO_PTHREAD completely breaks SDL2 (Which is used for display output in the demo)

LekKit:
By the way: It is likely a better idea to use `-sPROXY_TO_PTHREAD` here instead of `-sASYNCIFY`:

https://github.com/saeziae/homepage-rvvm-wasm/blob/84f99bc6896c65d5b6220edb09d5dd7f32345a11/build-rvvm.sh#L17

`ASYNCIFY` incurs a lot of CPU-side overhead which ideally we would like to avoid. The only reason why you wouldn't want to use `PROXY_TO_PTHREAD` is if you also want working SDL2 GUI and I've also implemented a solution for it (sorta) here: https://github.com/LekKit/test/blob/0bdee3508fbadf0cd46fca3cb1d6b47bf695e30a/wasm/emscripten-pty.js#L37

LekKit:
Also see 42447a9

This technically solves all of my previous obstacles towards implementing proper lazy disk image loading - i.e. expect you'll be able to pass arguments like `-image https://raw.githubusercontent.com/LekKit/test/refs/heads/master/wasm/rootfs.img` to RVVM WASM soon, and omit fully downloading the disk image upon page load (Instead acting more like a network attached storage).

`PROXY_TO_PTHREAD` is going to be the default from that point because I do not like when some threads have different semantics than the others.

PR #184, opened 2025-05-17
sound-hda: Correct stream management, working ALSA backend

epoll-reactor-2:
- HDA stream logic verified and works correctly.
- Sound propagated to host.
- **Note:** Volume knob is not handled to don't complicate device even more. Volume of the RVVM sound can be changed from the host.
- **Note:** No headphones and other non-standard output devices inside the VM. All sound should be played into default output device.

### DIFF
--- a/src/devices/alsa.c
+++ b/src/devices/alsa.c
@@ -1,0 +1,142 @@
+/*
+alsa.c - ALSA sound backend
+Copyright (C) 2025  David Korenchuk <github.com/epoll-reactor-2>
+
+This Source Code Form is subject to the terms of the Mozilla Public
+License, v. 2.0. If a copy of the MPL was not distributed with this
+file, You can obtain one at https://mozilla.org/MPL/2.0/.
+*/
+
+#ifdef USE_ALSA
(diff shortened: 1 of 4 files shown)


LekKit:
A few notes on future work which I might carry or assist if needed:

- We need an abstraction for audio devices / sinks / sources. I.e. it should be possible to swap out audio backends at runtime, as well as implementing one manually from RVVM public API. This might reside in `src/devices/sound_dev.h`, but the actual API design is yet to be decided upon. It would be great to be able to wrap sound device API into JNI (Java bindings), see https://github.com/LekKit/ScalarEvolution, as well as possible Android app port. 
- Implement more audio backends (winmm, SDL2, pulseaudio? I have some experience with winmm). Note that any audio backend that depends on an optional shared library should probe it via `dlib` dynamic linking facility, unless `USE_FULL_LINKING` is passed.

X547:
> Implement more audio backends (winmm, SDL2, pulseaudio? I have some experience with winmm)

\+ Native Haiku audio backend

PR #207, opened 2026-04-21
sound: Overhaul HDA stream path and ALSA backend

SolAstrius:
Six commits restoring the HDA output stream path end-to-end. Each is independently reviewable; they cluster into two themes — **stream worker correctness** and **host ALSA backend robustness** — with one API addition.

## Stream worker

1. **`sound-hda: Fix stream worker crashes, pacing, and codec format mismatch`** (`45d331c`)
   - `sound_hda_remove` halts the worker before `pci_func` teardown (prior: SIGSEGV on machine free).
   - Wall-clock pacing, derived per-stream from `stream->fmt` per HDA 7.3.3.10 (prior: `aplay` tripped `pcm_plugin.c:547` on non-blocking backends).
   - NULL-`dma` guard + `atomic_cas` on worker spawn.
   - Codec locked to 48 kHz mono 16-bit + defensive mono downmix in the worker.

2. **`sound-hda: Split worker lifetime from guest run-intent`** (`fe6683e`)
   Response to @epoll-reactor-2's review: the original CAS-on-`running` could race when guest stop+start interleaved with a worker mid-pacing-sleep. Split `worker_alive` (owned by worker) from `running` (guest intent). Drain entry is gated by CAS on `worker_alive`; worker re-claims the slot after its clear to cover the narrow missed-wakeup window. Exactly one worker ever executes `sound_hda_stream_drain` at a time.

3. **`sound-hda: Fix IRQ dispatch chain so period-elapsed wakes writers`** (`3c6cf25`)
   Three interdependent bugs silently dropped every output-stream IRQ before `snd_pcm_period_elapsed` ran. Streaming playback (mpg123) wrote a frame or two then reported `wrote only N of M (No error information?)`; file-backed `aplay` at low rates slipped through because the runtime buffer never filled. Fixes:
   - `INTSTS` was returning `1<<29`; Linux's `azx_interrupt` masks by stream index, expects `1<<1` for the single output stream.
   - BCIS (`SDnSTS` bit 2) wasn't being latched on IOC, so `snd_hdac_bus_handle_stream_irq` skipped the ack callback.
   - `WALLCLK` was stuck at 0; `intel.c:azx_position_ok` compares it against `period_wallclk * 2/3` and rejects every IRQ as bogus. Now backed by a 24 MHz counter off `rvtimer_clocksource`.

## Host ALSA backend

4. **`sound-alsa: Match HDA codec format and widen host PCM buffer`** (`f1d5a61`)
   - Open host PCM at 48 kHz to match the codec lock in commit 1 (prior: 192 kHz host vs 48 kHz codec → 4× fast playback).
   - 960-frame period / 3840-frame buffer (~20 ms / 80 ms at 48 kHz); 128 frames was too tight and caused PipeWire xruns under normal host scheduler jitter.

5. **`sound-alsa: Retry on xrun and drain short writes instead of dropping`** (`0e2b4ed`)
   `alsa_sound_write` called `snd_pcm_writei` exactly once; `-EPIPE` / `-ESTRPIPE` / `-EINTR` and positive short returns all silently dropped frames → audible clicks. Now loops until every frame is delivered with xrun/EINTR/EAGAIN retry handling, bounded retry count against a disconnected device.

## API

6. **`sound-hda: Add pluggable host-backend API (sound_hda_init_ex)`** (`bd0eb85`)
   Public `sound_hda_init_ex` / `_auto_ex` accepting a `void (*write_fn)(void*, void*, size_t)` + opaque user data. NULL preserves `USE_ALSA` behaviour byte-for-byte; non-NULL routes PCM through the caller. Existing `sound_hda_init` / `_auto` remain as one-line wrappers. Optional — happy to drop if your upcoming bindings API will supersede.

## Test plan

- [x] Linux x86_64 native build clean on GCC + Clang
- [x] All CI platforms green (Linux/Windows/macOS/FreeBSD/OpenBSD/Cosmo × arch matrix)
- [x] `aplay` no longer hits the `pcm_plugin.c:547` assertion
- [x] `apk add wget mpg123; mpg123 https://ice1.somafm.com/groovesalad-128` — sustained streaming playback works end-to-end in rvvm on a Linux host (also the main-branch test path for the sound stack)
- [x] Machine free during active stream no longer SIGSEGVs
- [x] Rapid guest stop+start cycles no longer produce overlapping drains
- [x] `USE_ALSA=0` builds unchanged; `USE_ALSA=1` with NULL `write_fn` byte-identical to pre-patch

### DIFF
--- a/src/devices/sound-hda.c
+++ b/src/devices/sound-hda.c
@@ -350,11 +350,9 @@ static void sound_hda_remove(rvvm_mmio_dev_t* dev)
     sound_hda_dev_t *hda = dev->data;
     if (hda != NULL) {
         atomic_store_uint32_relax(&hda->stream_output.running, 0);
-        // A BDL entry at 48 kHz × 2 bytes is typically 256-4096 bytes =
-        // ~3-43 ms of wall-clock per iteration under the new pacing loop.
-        // 20 ms is enough for the worker to notice the flag before the
-        // next iteration in the common (small-period) case; the worst-
-        // case long-period still exits at the top of the next loop.
+        // The HDA stream period is 128 frames at 192 kHz = ~667 µs wall.
+        // 20 ms of sleep is 30x that — comfortably long enough for the
+        // worker to finish its current iteration.
         sleep_ms(20);
     }
 }
@@ -796,15 +794,21 @@ static void *sound_hda_stream_worker(void *arg)
             uint32_t len  = bdle[1] & 0xFFFFFFFF;
             uint8_t  ioc  = bdle[1] >> 32 & 1;
 
+            // Dispatch PCM to the configured host-side backend. If no backend
+            // was installed at init time (neither a compile-time USE_ALSA nor
+            // a caller-supplied write_fn via sound_hda_init_ex), the HDA
+            // device enumerates on the PCI bus but this path is a no-op —
+            // the guest sees a working device and the LPIB counter advances
+            // so its driver doesn't stall, but PCM data is silently dropped.
+            //
             // Backend contract: always receives MONO 16-bit LE at the
             // stream's configured rate. If the guest driver configured a
             // multi-channel stream (Linux's HDA code likes stereo even
             // when the codec widget advertises mono capability), we
-            // downmix here by averaging channels — keeps the backend's
-            // job simple and the pacing math below stays 1:1 with the
-            // bytes it sees.
-#ifdef USE_ALSA
-            {
+            // downmix here by averaging channels. Keeps backends simple
+            // (they don't need to know channel count) and the pacing
+            // math below stays 1:1 with the bytes they see.
+            if (hda->subsystem.write != NULL) {
                 void *pcm = pci_get_dma_ptr(hda->pci_func, addr, len);
                 if (channels == 1 || pcm == NULL) {
                     hda->subsystem.write(&hda->subsystem, pcm, len);
@@ -831,15 +835,13 @@ static void *sound_hda_stream_worker(void *arg)
                         bytes_emitted += chunk_frames * 2;
                     }
                 } else {
-                    // Rare: non-16-bit multi-channel. Pass raw.
+                    // Rare: non-16-bit multi-channel. Pass raw; backends
+                    // that care can parse stream->fmt from the caller.
                     hda->subsystem.write(&hda->subsystem, pcm, len);
                 }
+            } else {
+                UNUSED(addr);
             }
-#else
-            // No compile-time backend: LPIB still advances (paced below)
-            // so the guest's HDA driver doesn't stall, but PCM is dropped.
-            UNUSED(addr);
-#endif
 
             // Wall-clock pacing. Compute the ideal elapsed time for the
             // bytes we've emitted so far and sleep the difference.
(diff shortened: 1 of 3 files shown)


src/devices/sound-hda.c
```diff
@@ -738,8 +888,16 @@ static void sound_hda_output_stream_ctl(sound_hda_dev_t *hda, uint32_t cmd)
...
-        atomic_store_uint32_relax(&stream->running, 1);
-        thread_create_task(sound_hda_stream_worker, hda);
+        // Only spawn a worker if one isn't already running. Without this
+        // guard, a guest that writes run=1 twice in quick succession (or
+        // a backend slow enough that one aplay ends before the worker
+        // exits and the next aplay starts) can leave two workers walking
+        // the same stream struct. They'd race on stream->lpib / running,
+        // and if one worker's cached dma pointer got stale from a guest
+        // BDL update in between, dereferencing it crashes.
+        if (atomic_cas_uint32(&stream->running, 0, 1)) {
```
epoll-reactor-2: I suspect it is possible to face the following situation:

1. `sound_hda_stream_worker()` sets `running` flag to 0
2. We still iterating over `total` data chunks in outer loop
3. `sound_hda_stream_worker()` has not yet completed
4. Meantime `sound_hda_output_stream_ctl()` swaps `running` flag to 1 and spawns a new thread
5. Worker thread will run twice at the same period of time

src/devices/sound-hda.c
```diff
@@ -692,41 +772,200 @@ static void sound_hda_codec_cmd(sound_hda_dev_t *hda, uint32_t cmd)
...
+    // missed-wakeup window: guest writes run=1 after we read running=0 but
+    // before we clear worker_alive — its spawn CAS sees worker_alive=1
+    // and skips, leaving no worker for a running stream. We catch that by
+    // re-reading running after the clear, and re-claim the slot to drain
+    // again. If a concurrent spawn already won the CAS, they own the next
+    // drain — we exit. Either way, exactly one worker runs at a time.
+    for (;;) {
+        sound_hda_stream_drain(hda);
+
+        atomic_store_uint32_relax(&stream->worker_alive, 0);
```
SolAstrius, line 961: Re. the (store wa=0 ... CAS wa 0→1) window here: the buffer stays race-free even if stream_ctl slips in between.

Drain entry is gated by a successful CAS on `worker_alive` — either this one at L965 (re-claim) or the spawn-path CAS at L993. CAS 0→1 is single-winner, so only one thread is ever past its successful CAS and executing `sound_hda_stream_drain`. If a concurrent spawn wins this window, its new task owns the next drain and this worker's CAS at L965 fails → returns without touching stream state. If this worker's CAS wins, the spawn's CAS at L993 fails → `thread_create_task` isn't called.

Ghost threads (spawned, lose their own CAS, return) can transiently exist during a toggle burst. They spend O(few atomics) and retire without entering drain — no shared state access, no race. The invariant `at most one thread in drain at any instant` holds in every interleaving.

src/devices/sound-hda.c
```diff
@@ -738,8 +977,22 @@ static void sound_hda_output_stream_ctl(sound_hda_dev_t *hda, uint32_t cmd)
...
 
     if (run) {
+        // Publish guest intent first, THEN gate the spawn on worker_alive.
+        // Ordering matters: a worker exiting its while-loop right now will
+        // clear worker_alive after its last running-check. If we set
+        // running=1 before it clears, our subsequent CAS will either win
+        // the slot (worker already gone) or lose it (worker still alive),
+        // and the worker's post-clear re-check of running catches the case
+        // where we lost the CAS but running=1 still needs a worker.
         atomic_store_uint32_relax(&stream->running, 1);
```
SolAstrius, line 987: The gap between this `store running=1` and the `CAS(worker_alive, 0, 1)` below is safe: `sound_hda_output_stream_ctl` runs under `hda->lock` (taken in `sound_hda_mmio_write` at L1006, released at L1125/1129). Any concurrent stream_ctl from another guest CPU serializes on the spinlock, so no other spawn or stop can observe the partial state.

The only concurrent thread is a worker — and it competes for the same `worker_alive` via the same CAS single-winner rule. Either a worker's post-drain CAS at L965 wins first (re-claim; this spawn's CAS at L993 then fails, no task created) or this spawn's CAS wins first (worker returns). Drain remains exclusive.

src/devices/sound-hda.c
```diff
@@ -738,8 +977,22 @@ static void sound_hda_output_stream_ctl(sound_hda_dev_t *hda, uint32_t cmd)
...
-        thread_create_task(sound_hda_stream_worker, hda);
+        // Gate on worker_alive (not running): running is guest intent and
+        // can flip repeatedly while a single worker drains a stream.
+        // worker_alive is owned by the worker — set here on spawn, cleared
+        // only by the worker at function return — so it stays 1 across the
+        // entire window where a second thread would trample stream state.
+        if (atomic_cas_uint32(&stream->worker_alive, 0, 1)) {
```
SolAstrius, line 993: Re. the suggestion to move `store running=1` (L987) into the worker thread: this would introduce a real race with the stop path.

```
T0  guest run=1 (under lock):
      CAS(wa, 0, 1) succeeds, wa=1
      thread_create_task queues worker, lock released
T1  guest run=0 (under lock):
      atomic_store(running, 0)            ← guest intent: STOP
      lock released. running=0, wa=1.
T2  worker scheduled → enters:
      atomic_store(running, 1)            ← overrides T1
      drain runs despite guest saying stop
```

The store's purpose isn't to gate the CAS (the CAS only reads wa). Its purpose is to publish guest intent *under the spinlock*, so a subsequent guest `run=0` store happens-after in lock order. The worker then observes `running=0` on drain's first while-check and exits cleanly. Moving the store to worker-entry loses that ordering — guest's T1 stop becomes overridable by T2 worker-start.

PR #219, opened 2026-06-15
Fix RISC-V FPU conformance issues

whensun:
This PR fixes RISC-V FPU conformance issues.

Changes include:

* Honor the `rm` field for `FCVT.D.L`.
* Set `NX` for inexact `FCVT.W.S` conversions.
* Adjust `FMSUB.S` qNaN invalid-flag behavior.

These changes were split out from the CSR/xRET fixes so that the FPU correctness and performance impact can be reviewed separately.

### DIFF
--- a/src/cpu/riscv_fpu.h
+++ b/src/cpu/riscv_fpu.h
@@ -164,12 +164,35 @@ static forceinline void riscv_emulate_f_fmsub(rvvm_hart_t* vm, const uint32_t in
 
     if (likely(riscv_fpu_is_enabled(vm) && riscv_fpu_rm_is_valid(rm))) {
         switch (bit_ext_u32(insn, 25, 2)) {
-            case 0x0: // fmsub.s
-                riscv_emit_s(vm, rds,
-                             fpu_fma32(riscv_view_s(vm, rs1), //
-                                       riscv_view_s(vm, rs2), //
-                                       fpu_neg32(riscv_view_s(vm, rs3))));
+            case 0x0: { // fmsub.s
+                fpu_f32_t a = riscv_view_s(vm, rs1);
+                fpu_f32_t b = riscv_view_s(vm, rs2);
+                fpu_f32_t c = riscv_view_s(vm, rs3);
+
+                uint32_t e_old = fpu_get_exceptions();
+                fpu_f32_t out  = fpu_fma32(a, b, fpu_neg32(c));
+                uint32_t e_new = fpu_get_exceptions();
+
+                bool c_is_qnan = fpu_is_nan32_soft(c) && !fpu_is_snan32_soft(c);
+
+                uint32_t ua = fpu_bit_f32_to_u32(a);
+                uint32_t ub = fpu_bit_f32_to_u32(b);
+                bool a_zero = (ua & 0x7FFFFFFFU) == 0;
+                bool b_zero = (ub & 0x7FFFFFFFU) == 0;
+                bool a_inf  = !fpu_is_finite32(a) && !fpu_is_nan32_soft(a);
+                bool b_inf  = !fpu_is_finite32(b) && !fpu_is_nan32_soft(b);
+                bool inf0_invalid = (a_inf && b_zero) || (b_inf && a_zero);
+
+                if (c_is_qnan && !inf0_invalid) {
+                    uint32_t raised = e_new & ~e_old;
+                    if (raised & FPU_LIB_FLAG_NV) {
+                        fpu_set_exceptions(e_new & ~FPU_LIB_FLAG_NV);
+                    }
+                }
+
+                riscv_emit_s(vm, rds, out);
                 return;
+            }
             case 0x1: // fmsub.d
                 riscv_emit_d(vm, rds,
                              fpu_fma64(riscv_view_d(vm, rs1), //
--- a/src/cpu/riscv_fpu.h
+++ b/src/cpu/riscv_fpu.h
@@ -164,35 +164,12 @@ static forceinline void riscv_emulate_f_fmsub(rvvm_hart_t* vm, const uint32_t in
 
     if (likely(riscv_fpu_is_enabled(vm) && riscv_fpu_rm_is_valid(rm))) {
         switch (bit_ext_u32(insn, 25, 2)) {
-            case 0x0: { // fmsub.s
-                fpu_f32_t a = riscv_view_s(vm, rs1);
-                fpu_f32_t b = riscv_view_s(vm, rs2);
-                fpu_f32_t c = riscv_view_s(vm, rs3);
-
-                uint32_t e_old = fpu_get_exceptions();
-                fpu_f32_t out  = fpu_fma32(a, b, fpu_neg32(c));
-                uint32_t e_new = fpu_get_exceptions();
-
-                bool c_is_qnan = fpu_is_nan32_soft(c) && !fpu_is_snan32_soft(c);
-
-                uint32_t ua = fpu_bit_f32_to_u32(a);
-                uint32_t ub = fpu_bit_f32_to_u32(b);
-                bool a_zero = (ua & 0x7FFFFFFFU) == 0;
-                bool b_zero = (ub & 0x7FFFFFFFU) == 0;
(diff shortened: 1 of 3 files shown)


src/cpu/riscv_fpu.h
```diff
@@ -164,12 +164,35 @@ static forceinline void riscv_emulate_f_fmsub(rvvm_hart_t* vm, const uint32_t in
...
+
+                uint32_t ua = fpu_bit_f32_to_u32(a);
+                uint32_t ub = fpu_bit_f32_to_u32(b);
+                bool a_zero = (ua & 0x7FFFFFFFU) == 0;
+                bool b_zero = (ub & 0x7FFFFFFFU) == 0;
+                bool a_inf  = !fpu_is_finite32(a) && !fpu_is_nan32_soft(a);
+                bool b_inf  = !fpu_is_finite32(b) && !fpu_is_nan32_soft(b);
+                bool inf0_invalid = (a_inf && b_zero) || (b_inf && a_zero);
+
+                if (c_is_qnan && !inf0_invalid) {
```
SolAstrius: This is a regression. You took code that was correct on every mainstream build and broke it. That is worse than the bug you think you're fixing.

IEEE 754-2008, 7.2 list item (a): the invalid operation exception is signaled for "any general-computational or signaling-computational operation on a signaling NaN." An FMA is general-computational. sNaN in, NV out. Not optional.

Now read item (c), two lines down: for fusedMultiplyAdd(0, ∞, c) "if c is a quiet NaN then it is implementation defined whether the invalid operation exception is signaled." The standard already carved out the quiet-NaN case you were fretting over and left it to the implementation - and RISC-V then mandates it anyway (6.1.6). So the qNaN part was never the hard part.

The hard part was not trampling item (a) while you were in there. But you inspect the addend and nothing else - never rs1, never rs2. Feed an sNaN multiplicand with a qNaN in c and your patch clears the NV the hardware correctly raised. The commit titled "Fix FMSUB quiet NaN invalid flag handling" mishandles the invalid flag for signaling NaNs.

Check a and b for sNaN before you go deleting flags you didn't raise and don't understand.

src/util/fpu_lib.h
```diff
@@ -1090,6 +1090,76 @@ static forceinline fpu_f64_t fpu_fcvt_i64_to_f64(int64_t i)
...
+
+        if (increment) {
+            ++sig;
+            if (unlikely(sig == (1ULL << 53))) {
+                sig >>= 1;
+                ++exp;
+            }
+        }
+
+        sig <<= 52 - exp + shift;
```
SolAstrius: Walk me through what you think this line does, because it does nothing you want.

Normal path: shift by zero. A no-op.
Carry path: exp got bumped, so 52 - exp + shift is -1, and since it's all unsigned that's a shift by 4294967295.

That is undefined behavior. The most basic kind - shift count larger than the type. The only reason it hasn't blown up in your face is luck: a carry always leaves a zero mantissa, and the CPU silently masks your absurd shift count down to something harmless. You shipped UB and the silicon happened to cover for you.

Delete the line. That's the whole fix. Think about how a no-op-or-UB line ended up in code you sent for review.

src/cpu/riscv_fpu.h
```diff
@@ -164,12 +164,35 @@ static forceinline void riscv_emulate_f_fmsub(rvvm_hart_t* vm, const uint32_t in
 
     if (likely(riscv_fpu_is_enabled(vm) && riscv_fpu_rm_is_valid(rm))) {
         switch (bit_ext_u32(insn, 25, 2)) {
-            case 0x0: // fmsub.s
-                riscv_emit_s(vm, rds,
-                             fpu_fma32(riscv_view_s(vm, rs1), //
-                                       riscv_view_s(vm, rs2), //
-                                       fpu_neg32(riscv_view_s(vm, rs3))));
+            case 0x0: { // fmsub.s
```
SolAstrius: You clearly read this -

  RISC-V Unprivileged ISA, 6.1.6 (Single-Precision Floating-Point Computational Instructions), p.132: "The fused multiply-add instructions must set the invalid operation exception flag when the multiplicands are ∞ and zero, even when the addend is a quiet NaN."

- and then reimplemented half of it, by hand, badly, inside the instruction decoder. Why is any of this here? The bug is that your soft-fenv FMA raises NV for every NaN result. Fix it once, where the FMA actually lives, and every fmadd/fmsub/fnmadd/fnmsub in single and double comes out correct for nothing.

Instead you special-cased one opcode, got the sNaN handling wrong (see the comment on the clear block below), and left fmadd.s, fnmsub.s, fnmadd.s and every .d variant exactly as broken as you found them. This isn't a fix. It's three lines of patch and ten more of new bug.

src/util/fpu_lib.h
```diff
@@ -1333,10 +1403,21 @@ static forceinline uint32_t fpu_round_f64_to_u32(fpu_f64_t d, uint32_t mode)
...
+    fpu_f32_t rounded = f;
+
+    if (likely(mode != FPU_LIB_ROUND_TZ)) {
+        rounded = fpu_round_f32_internal(f, mode);
     }
-    return fpu_fcvt_f32_to_i32(fpu_round_f32_internal(f, mode));
+
+    int32_t ret = fpu_fcvt_f32_to_i32(rounded);
+
+    if (likely(!(fpu_get_exceptions() & FPU_LIB_FLAG_NV))) {
```
SolAstrius, line 1528: This one's correct, and it even matches the book:

  RISC-V Unprivileged ISA, 6.1.7 (Single-Precision Floating-Point Conversion and Move Instructions), p.133: "All floating-point conversion instructions set the Inexact exception flag if the rounded result differs from the operand value and the Invalid exception flag is not set."

Read the first word. ALL. Not "fcvt.w.s." All of them.

So you fixed one and stopped. fcvt.l.s, fcvt.w.d, fcvt.l.d and the unsigned variants have the identical missing-NX bug, sitting in the same file, inches away, untouched. Finish the job, or at least be honest in the commit message that w.s is the only one you did.

SolAstrius:
NAK.

Four changes, and the one that matters is a regression. The fmsub.s "fix" deletes the invalid flag on sNaN multiplicands - behavior that was correct on every mainstream build, now broken. A patch that turns correct code into wrong code does not merge. That's the end of the discussion, not the start of it.

And the spec makes it worse, because it spells out exactly which case is which. IEEE 754-2008, 7.2:

  (a) invalid is signaled for "any general-computational or signaling-computational operation on a signaling NaN" - mandatory, no wiggle room.
  (c) for fusedMultiplyAdd(0, infinity, c), "if c is a quiet NaN then it is implementation defined whether the invalid operation exception is signaled."

So the standard already made the quiet-NaN case you were chasing implementation-defined (and RISC-V 6.1.6 then mandates it). That part was free. The signaling-NaN case in (a) is the one that is NOT optional - and that is the exact one you broke. You trampled the mandatory rule to fix the discretionary one.

The rest, briefly:

- The i64->f64 rounding carry path is undefined behavior: a shift by (unsigned)-1. It only produces the right answer because the hardware masks the shift count for you and the carry happens to leave a zero mantissa. Luck is not a correctness argument. Delete the line.

- The qNaN/NV handling has no business being open-coded in the instruction decoder. The actual bug is your soft-fenv FMA raising NV for every NaN result. Fix it there, once, and fmadd/fmsub/fnmadd/fnmsub in both precisions come out correct. As written you patched one opcode, got it wrong, and left the other seven exactly as broken as they were.

- The inexact fix is correct, and it's the only thing in here worth keeping - and you did it for fcvt.w.s alone while leaving the identical missing-NX bug in fcvt.l.s, fcvt.w.d, fcvt.l.d and the unsigned variants sitting right next to it.

So: keep the inexact work and finish it for all the conversions. Drop the UB line. Move the FMA fix into the FMA and make it check for signaling NaNs - item (a), not item (c). Bin the rest.

And stop bundling FPU flag changes you plainly never tested against the spec with the one thing you got right. One correct patch beats four where three are broken.

whensun:
Thanks for the detailed review! I will push an updated version to fix these things.

whensun:
Updated the FPU PR based on the review.

Please take another look.

PR #229, opened 2026-06-18
gui: map evdev keycodes to HID directly, bypassing host keyboard layout

SolAstrius:
Both the Wayland and X11 backends previously delivered HID usage codes based on the *keysym* for each physical key. `xkb_state_key_get_one_sym()` (Wayland) and `XGetKeyboardMapping()`'s slot 0 (X11) both apply the user's loaded keyboard layout, so a host running Workman / Dvorak / Colemak / AZERTY would feed the emulator layout-translated letters rather than physical key positions.

For an emulator pretending to be hardware with its own keyboard — or its own layout assumptions, e.g. the ZX Spectrum expects USB-HID-style positions for its 8×5 matrix lookup — that's the wrong abstraction. Layout translation should happen *inside* the emulated machine, not on the host side.

This routes both backends through a direct evdev-keycode → HID usage table:

- **Wayland**: `wl_keyboard.key` delivers `key` as the evdev keycode. Map directly via `wayland_evdev_keycode_to_hid()`.
- **X11**: `xkey.keycode` is evdev+8 on Linux X11. Subtract 8 and look up in the same table. Falls through to the existing xkb-keysym path for non-Linux X11 servers (where the offset isn't 8) or for keycodes outside the standard evdev range.

Result: the emulator sees the physical key the user pressed, regardless of host keyboard layout. On Wayland the keysym path is now fully unreachable, so it's removed; on X11 it's retained as the non-Linux fallback.

### DIFF
--- a/src/gui/wayland_window.c
+++ b/src/gui/wayland_window.c
@@ -201,180 +201,86 @@ static spinlock_t    wl_lock    = ZERO_INIT;
 static thread_ctx_t* wl_thread  = NULL;
 
 /*
- * XKB to HID key mapping
- */
+ * Linux evdev keycode → HID usage
+ *
+ * Wayland's wl_keyboard.key event delivers `key` as a Linux input
+ * event keycode (evdev). That's a STABLE physical-key identifier —
+ * it doesn't depend on which keyboard layout the user has loaded.
+ * Going via xkb_state_key_get_one_sym() instead would apply the user's
+ * current layout (Workman, Dvorak, Colemak, AZERTY, ...) and the
+ * emulator would receive layout-translated letters rather than
+ * physical key positions. For a hardware emulator that wants its own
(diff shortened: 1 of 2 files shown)


LekKit:
I am not sure this is the correct way, at least not by default:
- This would make X11/Wayland backends incompatible with SDL/Win32/Cocoa, where there is no way to get a native hardware key like this.
- In some cases, this would be the opposite of what user may have wanted: Now they need to configure each guest to use their preferred layout, instead of the VM recognizing it from the get-go
- Personally, I find the idea "user configures preferred layout on the host, and the guest is expected to just use QWERTY" more appealing, especially given the possibility of software-simulated input.
- The Wayland backend is _not_ Linux-only, even if it is currently only built on Linux by default. Various *BSDs are adopting Wayland, and Haiku OS has Wayland support, translating into its native compositor. I am not sure all of those systems support evdev and evdev keycodes.

PR #235, opened 2026-06-18
riscv_fpu: honor static RMM rounding-mode field

SolAstrius:
## Summary

Extends #233 so that RMM (IEEE 754 `roundTiesToAway`) is honored when it is selected by the **static `rm` field** of an instruction (e.g. `fadd.s …,rmm`), not only by the dynamic `frm` CSR.

> **Draft / stacked on #233.** This branch builds on the #233 commit, so until that merges this PR shows both commits; the new work is the top commit, *"riscv_fpu: honor static RMM rounding-mode field."* Please review/merge #233 first.

## Background

#233 fixes RMM but only triggers on `frm == RMM` (dynamic rounding). The RISC-V spec also allows RMM to be encoded statically in the instruction's `rm` field; RVVM previously ignored a static `rm` for arithmetic ops entirely (only `frm` drove the host mode). So `fadd.s rd,rs1,rs2,rmm` under the default `frm` was computed as round-to-nearest-even.

## Change

- Detect RMM from either source: `rm == 0b100` (static) **or** `rm == DYN && frm == RMM` (dynamic). `funct3 == 0b100` only ever appears on rounding-capable ops, so the static check can't misfire on `fsgnj`/`fcmp`/`fclass`/etc.
- The error-free transforms (TwoSum / TwoProduct) used for the ties-away fixup require the base op to be in round-to-nearest. A static `,rmm` under a *directed* `frm` is the only case where the host (which tracks `frm`) isn't already nearest, so the fixup helpers wrap the op in a **self-eliminating** guard: it reads the host mode and only switches to RNE (restoring after) when it isn't already RNE. For the common cases (dynamic RMM, or static `,rmm` under the default `frm`) it's a no-op, so the hot path is untouched.

## Validation

- The regression harness now runs **every** vector twice — dynamic `frm=RMM` and static `,rmm` — plus a dedicated check that a directed `frm` (e.g. RTZ) survives a static-`,rmm` op (host mode restored). All pass (3006 vector-checks + restore), MPFR-oracle expected values, f32+f64, incl. subnormals/boundary.
- No performance regression: this is implemented in a single dispatch function (an earlier wrapper-based version that added a per-op call was reworked out); FP throughput matches baseline.

## Out of scope (same as #233)

- Static `rm` for the *other* directed modes (rtz/rdn/rup) on arithmetic is still not honored — only RMM is addressed here.
- `fcvt`-to-integer under RMM still rounds ties-to-even.

### DIFF
--- a/src/cpu/riscv_fpu.c
+++ b/src/cpu/riscv_fpu.c
@@ -57,10 +57,11 @@ static const uint32_t riscv_fli_table[32] = {
 /*
  * RMM (round to nearest, ties to max magnitude) == IEEE 754 roundTiesToAway.
  *
- * The host FPU has no such mode, so when frm == RMM the host is left in
- * round-to-nearest-even (see fpu_set_rounding_mode()). RNE and roundTiesToAway
- * produce identical results EXCEPT on an exact halfway tie, where RNE rounds to
- * even and roundTiesToAway rounds to the larger-magnitude neighbour.
+ * The host FPU has no such mode. RMM may be requested via the dynamic frm CSR or
+ * a static instruction rm field; either way riscv_emulate_f_opc_op() runs the op
+ * in round-to-nearest. RNE and roundTiesToAway produce identical results EXCEPT
+ * on an exact halfway tie, where RNE rounds to even and roundTiesToAway rounds to
+ * the larger-magnitude neighbour.
  *
  * So we compute the op in RNE, recover the EXACT rounding error via the library's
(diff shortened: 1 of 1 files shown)


SolAstrius:
Superseded by #233, which now carries static rounding-mode support as its own commit ("honor the static rounding-mode field") and generalizes it to all static modes (rtz/rdn/rup/rmm), not just RMM.

PR #236, opened 2026-06-19
riscv_fpu: canonicalize NaN results of arithmetic and FMA ops

SolAstrius:
## Summary

RISC-V mandates that any operation producing a NaN result returns the **canonical** NaN (`0x7fc00000` / `0x7ff8000000000000`), regardless of the input NaN payloads. RVVM's `fadd/fmul/fdiv/fsqrt` and the whole FMA family used `riscv_emit_*` (which only NaN-boxes), so they propagated the **host** NaN payload instead. `fsub` already used the canonicalizing `riscv_write_*` — this just makes the rest consistent.

Switches the NaN-producing arithmetic ops (`fadd/fmul/fdiv/fsqrt`, f32+f64) and the FMA family (`fmadd/fmsub/fnmadd/fnmsub`, f32+f64) from `riscv_emit_*` to `riscv_write_*`. Sign-manipulation (`fsgnj*`), `fmin/fmax` (which canonicalize internally), `fcvt`, and `fmv`/`fli` are deliberately left on `riscv_emit_*` — they must not canonicalize.

> **Stacked on #233.** This branches off #233 (`fix/rmm-rounding`), so until that merges this PR shows its three commits plus the one here (*"canonicalize NaN results of arithmetic and FMA ops"*). Please take #233 first.

## Validation

- The RMM MPFR harness (`rvvm-hal/examples/rmm-test`) still passes 3006/3006 — no value or rounding regression.
- RISC-V arch-test (ACT4 / Spike reference): on top of #233, **F 22→34, D 32→44** (the `fadd/fmul/fdiv` failures whose first mismatch was a non-canonical NaN result — e.g. `0x7fffffff` where the spec wants `0x7fc00000`). Combined I/M/F/D = **142/260**.

## Out of scope (separate follow-ups, pre-existing)

These are mapped per-instruction by the arch-test diagnostic but not addressed here:
- `fcvt`-to-int inexact-flag handling (spurious `NX`, incl. alongside `NV`).
- The FMA family's **rounding-mode** handling (it currently ignores `rm` entirely) and invalid-flag corner cases — NaN canonicalization for FMA is the only FMA gap closed here.
- Mal-boxed narrow (`.s`) operands must read as the canonical NaN.
- `fsqrt(-0.0)` should return `-0.0`.

### DIFF
--- a/src/cpu/riscv_fpu.c
+++ b/src/cpu/riscv_fpu.c
@@ -54,43 +54,160 @@ static const uint32_t riscv_fli_table[32] = {
     0x7FC00000UL, // Canonical NaN
 };
 
-static void riscv_prepare_rmm(rvvm_hart_t* vm, const uint32_t insn, const size_t rs1, const size_t rs2)
+/*
+ * RMM (round to nearest, ties to max magnitude) == IEEE 754 roundTiesToAway.
+ *
+ * The host FPU has no such mode, so when frm == RMM the host is left in
+ * round-to-nearest-even (see fpu_set_rounding_mode()). RNE and roundTiesToAway
+ * produce identical results EXCEPT on an exact halfway tie, where RNE rounds to
+ * even and roundTiesToAway rounds to the larger-magnitude neighbour.
+ *
+ * So we compute the op in RNE, recover the EXACT rounding error via the library's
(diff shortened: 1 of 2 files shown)


SolAstrius:

PR #237, opened 2026-06-19
riscv_fpu: honor the rounding mode in the FMA ops

SolAstrius:
## Summary

Makes the fused-multiply-add family honor the rounding mode. Two issues:

1. **The rounding mode was ignored unless it came from the dynamic `frm` CSR.** `fpu_fma` rounds in the host mode (which only tracks `frm`), so a **static** `rm` field on an FMA (`fmadd.s …,rtz` etc.) was computed in whatever `frm` left set, and RMM had no host mode at all. A small wrapper (`riscv_fma_round_f32/64`, mirroring the OP-FP dispatch in #233) applies the effective mode around the op: static host-native modes are set/restored, RMM runs in RNE.
2. **`fnmadd` rounded the wrong value.** It computed `-(fma(a,b,c))` — rounding `a*b+c` and *then* negating, which flips the rounding **direction** under the directed modes. `fnmadd = -(rs1*rs2) - rs3` must be `fma(-a, b, -c)` so the single rounding sees the correctly-signed result. (`fmsub`/`fnmsub` already negated operands, not the result — only `fnmadd` was wrong.)

> **Stacked on #236 → #233.** Until those merge this shows their commits plus the one here. Independent in logic; based on #236 only because both touch the FMA handlers.

## Validation

- RMM MPFR harness: still 3006/3006 — no regression.
- RISC-V arch-test (ACT4 / Spike): on top of #233+#236, **F 34→54, D 44→69** (+45). Combined I/M/F/D = **187/260** (baseline `staging` is 101).
- `__builtin_fma` (used on aarch64/riscv/ppc) respects the host mode for every rounding mode, so the wrapper is exact there; the software fallback rounds via round-to-odd, which is also mode-correct.

## Out of scope (pre-existing, separate)

- **RMM exact ties in FMA** (~10 tests): an exact FMA halfway case still rounds to-nearest-even, not away — the same ties-away gap shared with the rest of the FP path; the exact FMA residual needs an error-free FMA, deferred.
- **Spurious `UF`** on a few near-underflow FMAs (`0x03` vs `0x01`): an FMA exception-flag bug, not a rounding bug.

### DIFF
--- a/src/cpu/riscv_fpu.c
+++ b/src/cpu/riscv_fpu.c
@@ -54,43 +54,160 @@ static const uint32_t riscv_fli_table[32] = {
     0x7FC00000UL, // Canonical NaN
 };
 
-static void riscv_prepare_rmm(rvvm_hart_t* vm, const uint32_t insn, const size_t rs1, const size_t rs2)
+/*
+ * RMM (round to nearest, ties to max magnitude) == IEEE 754 roundTiesToAway.
+ *
+ * The host FPU has no such mode, so when frm == RMM the host is left in
+ * round-to-nearest-even (see fpu_set_rounding_mode()). RNE and roundTiesToAway
+ * produce identical results EXCEPT on an exact halfway tie, where RNE rounds to
+ * even and roundTiesToAway rounds to the larger-magnitude neighbour.
+ *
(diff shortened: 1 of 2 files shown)


SolAstrius:

PR #239, opened 2026-06-19
riscv_fpu: IEEE 754 rounding and exception conformance for the scalar FPU

SolAstrius:
## Summary

Bring RVVM's scalar F/D floating-point rounding and exception behaviour up to IEEE 754 / RISC-V conformance.

This began as the [#204](https://github.com/LekKit/RVVM/issues/204) RMM (`roundTiesToAway`) fix and grew to close the rounding and flag gaps the conformance suite exposed.

## What's fixed

- **RMM rounding** computed via exact error-free transforms, and the instruction's static `rm` field honoured (not just `frm`) for all arithmetic ops.
- **FP-op dispatch** no longer size-optimised (`func_opt_size` dropped) — roughly 2x interpreter FP throughput at zero size cost, since the FP path is not JIT'd.
- **The FMA family**: rounding mode honoured (including the `fnmadd` operand-negation fix), RMM exact ties via an error-free FMA, and the underflow flag set by true IEEE *after-rounding* tininess.
- **fmul** ties-away near the underflow boundary, where the Dekker product error itself underflows.
- **sqrt(-0.0)** returns `-0.0` without raising invalid.
- A spurious **`INEXACT`** flag leaked by integer rounding, removed.

Everything stays within the existing `fpu_lib` wrapper discipline — no raw host FP, no host-fenv-only tricks — so it remains compatible with a software-fenv / jittable softfloat path.

## Validation

- **MPFR-oracle bare-metal harness** (rvvm-hal `examples/rmm-test`): `43652/0` over the five OP-FP ops and the four FMA ops, each under both dynamic `frm=RMM` and the static `,rmm` suffix, including subnormal and underflow-boundary ties.
- Together with the canonical-NaN changes (#240), **riscv-arch-test (ACT4, Spike reference)** F/D/I/M reaches `260/260`.

## Notes

Commits are split one-fix-per-commit for review. No functional change outside `src/cpu/riscv_fpu.{c,h}` and `src/util/fpu_lib.{c,h}`.

### DIFF
--- a/src/cpu/riscv_fpu.c
+++ b/src/cpu/riscv_fpu.c
@@ -68,7 +68,8 @@ static const uint32_t riscv_fli_table[32] = {
  * (The previous implementation rounded toward +/-inf unconditionally, which is
  * correct on ties but wrong for every inexact non-tie. See issue #204.)
  *
- * fadd/fsub/fmul use riscv_rmm_apply (below). fsqrt never needs a fixup: a square
+ * fadd/fsub/fmul use riscv_rmm_apply (in riscv_fpu.h, shared with the FMA family).
+ * fsqrt never needs a fixup: a square
  * root is irrational unless exact, and its result is always normal (the square
  * root of even the smallest subnormal is ~2^-75), so RNE == roundTiesToAway. A
  * *normally-rounded* quotient is likewise never an exact halfway case — but a
@@ -81,40 +82,6 @@ static const uint32_t riscv_fli_table[32] = {
  * spurious exceptions (inf-inf -> NV, near-FLT_MAX -> OF) that must not leak into
  * fflags. The genuine flags are already set by the base op.
  */
-static forceinline fpu_f32_t riscv_rmm_apply_f32(fpu_f32_t n, fpu_f32_t err)
-{
-    const uint32_t un = fpu_bit_f32_to_u32(n);
-    const uint32_t ue = fpu_bit_f32_to_u32(err);
-    // Skip exact results (err == +/-0) and errors pointing toward zero: in both
-    // cases RNE already delivers the roundTiesToAway value.
-    if ((ue << 1) == 0 || (un >> 31) != (ue >> 31)) {
-        return n;
-    }
-    // n + 1 ULP toward larger magnitude, and the exact gap to it.
-    const fpu_f32_t away    = fpu_bit_u32_to_f32(un + 1);
-    const fpu_f32_t spacing = fpu_sub32(away, n);  // exact: adjacent floats
-    // Tie iff the exact result is the midpoint, i.e. 2*err == spacing.
-    if (fpu_is_bit_equal32(fpu_add32(err, err), spacing)) {
-        return away;
-    }
-    return n;
-}
-
-static forceinline fpu_f64_t riscv_rmm_apply_f64(fpu_f64_t n, fpu_f64_t err)
-{
-    const uint64_t un = fpu_bit_f64_to_u64(n);
-    const uint64_t ue = fpu_bit_f64_to_u64(err);
-    if ((ue << 1) == 0 || (un >> 63) != (ue >> 63)) {
-        return n;
-    }
-    const fpu_f64_t away    = fpu_bit_u64_to_f64(un + 1);
-    const fpu_f64_t spacing = fpu_sub64(away, n);
-    if (fpu_is_bit_equal64(fpu_add64(err, err), spacing)) {
-        return away;
-    }
-    return n;
-}
-
 // roundTiesToAway fixup for fadd/fsub (fsub passes b negated). Flag-isolated.
 static forceinline fpu_f32_t riscv_rmm_add_f32(fpu_f32_t n, fpu_f32_t a, fpu_f32_t b)
 {
--- a/src/cpu/riscv_fpu.c
+++ b/src/cpu/riscv_fpu.c
@@ -105,14 +105,62 @@ static forceinline fpu_f64_t riscv_rmm_add_f64(fpu_f64_t n, fpu_f64_t a, fpu_f64
     return r;
 }
 
-// roundTiesToAway fixup for fmul. Flag-isolated.
+/*
+ * Dekker's product error (fpu_mul_error) loses bits once the product error drops
+ * below the smallest subnormal -- not only for a subnormal result, but throughout
+ * the smallest normal binades (the half-ULP ~2^(e-53) underflows for e <= -969).
+ * Those results need an exact tie test.
+ *
+ * For f32 the exact product widens losslessly into f64 (48 <= 53 bits) and the
+ * away-side midpoint is exact there (<= 25 bits, down to 2^-150), so the op is a
+ * tie iff a*b == m. This is exact for every f32 result, so fmul.s always uses it.
+ *
+ * For f64 there is no wider type, so scale both operands by 2^512 to lift the
+ * product into a binade where the fma-based TwoProduct is exact (a small product
+ * bounds |a|,|b|, so a*2^512 cannot overflow). The midpoint of two normals needs
+ * one bit more than the format holds, so rather than form it we test the residual:
+ * the op is a tie iff 2*(a*b - n) == ULP, evaluated in the scaled domain with an
+ * exact TwoSum cancellation. Mirrors the scaling in riscv_rmm_div_apply_f64.
+ */
+static forceinline fpu_f32_t riscv_rmm_mul_exact_f32(fpu_f32_t n, fpu_f32_t a, fpu_f32_t b)
+{
+    const fpu_f32_t away = fpu_bit_u32_to_f32(fpu_bit_f32_to_u32(n) + 1);
+    const fpu_f64_t half = fpu_bit_u64_to_f64(0x3FE0000000000000ULL);  // 0.5
+    const fpu_f64_t dn   = fpu_fcvt_f32_to_f64(n);
+    const fpu_f64_t dab  = fpu_mul64(fpu_fcvt_f32_to_f64(a), fpu_fcvt_f32_to_f64(b));  // exact
+    const fpu_f64_t m    = fpu_add64(dn, fpu_mul64(fpu_sub64(fpu_fcvt_f32_to_f64(away), dn), half));
+    return ((fpu_bit_f64_to_u64(fpu_sub64(dab, m)) << 1) == 0) ? away : n;  // tie iff a*b == m
+}
+
+static forceinline fpu_f64_t riscv_rmm_mul_small_f64(fpu_f64_t n, fpu_f64_t a, fpu_f64_t b)
+{
+    const fpu_f64_t S    = fpu_bit_u64_to_f64(0x5FF0000000000000ULL);  // 2^512
+    const fpu_f64_t as   = fpu_mul64(a, S), bs = fpu_mul64(b, S);      // exact: |a|,|b| bounded
+    const fpu_f64_t ps   = fpu_mul64(as, bs);                          // a*b * 2^1024, normal
+    const fpu_f64_t pe   = fpu_fma64(as, bs, fpu_neg64(ps));           // exact: a*b*2^1024 = ps + pe
+    const fpu_f64_t away = fpu_bit_u64_to_f64(fpu_bit_f64_to_u64(n) + 1);
+    const fpu_f64_t ns   = fpu_mul64(fpu_mul64(n, S), S);              // n * 2^1024, exact
+    const fpu_f64_t gaps = fpu_mul64(fpu_mul64(fpu_sub64(away, n), S), S);  // ULP * 2^1024
+    // residual (a*b - n)*2^1024 = dr + pe (dr exact, Sterbenz). Tie iff 2*(a*b-n)
+    // == ULP, i.e. 2*dr + 2*pe == gaps, via an exact TwoSum(2*dr, -gaps).
+    const fpu_f64_t dr   = fpu_sub64(ps, ns);
+    const fpu_f64_t td   = fpu_add64(dr, dr), tp = fpu_add64(pe, pe), ng = fpu_neg64(gaps);
+    const fpu_f64_t vh   = fpu_add64(td, ng);
+    const fpu_f64_t vl   = fpu_add_error64(vh, td, ng);               // exact: (td - gaps) = vh + vl
+    return (((fpu_bit_f64_to_u64(vl) << 1) == 0) && ((fpu_bit_f64_to_u64(fpu_add64(vh, tp)) << 1) == 0))
+               ? away : n;                                            // tie iff td - gaps == -2*pe
+}
+
+// roundTiesToAway fixup for fmul. Flag-isolated. fmul.s always uses the exact f64
+// widening; fmul.d uses the exact scaled test for small results (where the product
+// error underflows) and the cheaper Dekker product error otherwise.
 static forceinline fpu_f32_t riscv_rmm_mul_f32(fpu_f32_t n, fpu_f32_t a, fpu_f32_t b)
 {
     if (unlikely(!fpu_is_finite32(n))) {
         return n;
     }
     const uint32_t exc = fpu_get_exceptions();
-    const fpu_f32_t r  = riscv_rmm_apply_f32(n, fpu_mul_error32(n, a, b));
+    const fpu_f32_t r  = riscv_rmm_mul_exact_f32(n, a, b);
     fpu_set_exceptions(exc);
     return r;
 }
@@ -123,7 +171,11 @@ static forceinline fpu_f64_t riscv_rmm_mul_f64(fpu_f64_t n, fpu_f64_t a, fpu_f64
         return n;
     }
     const uint32_t exc = fpu_get_exceptions();
-    const fpu_f64_t r  = riscv_rmm_apply_f64(n, fpu_mul_error64(n, a, b));
+    // |n| >= 2^-959 (exp field >= 64): the product error stays representable, so
+    // the cheap Dekker path is exact. Below that, use the scaled residual test.
+    const fpu_f64_t r  = (((fpu_bit_f64_to_u64(n) >> 52) & 0x7FFU) >= 64)
+                             ? riscv_rmm_apply_f64(n, fpu_mul_error64(n, a, b))
+                             : riscv_rmm_mul_small_f64(n, a, b);
     fpu_set_exceptions(exc);
     return r;
 }
--- a/src/cpu/riscv_fpu.c
+++ b/src/cpu/riscv_fpu.c
@@ -54,43 +54,160 @@ static const uint32_t riscv_fli_table[32] = {
     0x7FC00000UL, // Canonical NaN
 };
 
-static void riscv_prepare_rmm(rvvm_hart_t* vm, const uint32_t insn, const size_t rs1, const size_t rs2)
+/*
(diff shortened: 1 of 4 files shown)


SolAstrius:
@purplesyringa please help us with your wisdom 🙏

src/cpu/riscv_fpu.c
```diff
@@ -406,4 +561,32 @@ slow_path func_opt_size void riscv_emulate_f_opc_op(rvvm_hart_t* vm, const uint3
...
+    const uint32_t frm = vm->csr.fcsr >> 5;
+    // Effective rounding mode: a static rm field overrides the dynamic frm CSR.
+    const uint32_t eff = (rm == 0x07) ? frm : rm;
+    const bool     rmm = (eff == 0x04);
+    // Override the host mode when synthesizing RMM (run sub-ops in RNE) or when a
+    // static rm field selects a host-native mode other than the one frm left set.
+    if (unlikely(rmm || (rm != 0x07 && eff != frm))) {
+        const uint32_t host = fpu_get_rounding_mode();
+        fpu_set_rounding_mode(rmm ? FPU_LIB_ROUND_NE : eff);
+        riscv_emulate_f_opc_op_impl(vm, insn, rmm);
```
purplesyringa, line 585: Does it actually make sense to handle RMM in `riscv_fpu`? This feels like something that is better integrated into `fpu_lib`, where the rest of math-heavy logic lives. This arbitrary separation confused me.

src/util/fpu_lib.h
```diff
@@ -1071,12 +1071,12 @@ static forceinline fpu_f32_t fpu_sqrt32(fpu_f32_t f)
 #endif
         return ret;
     }
-    if (fpu_is_negative32(f)) {
-        // Raise invalid flag, return canonical NaN
+    if (fpu_is_negative32(f) && (fpu_bit_f32_to_u32(f) << 1) != 0) {
```
purplesyringa, line 1074: Maybe add a `#define FPU_LIB_FPxx_NEGATIVE_ZERO` to `fpu_lib` to make the intent of this comparison clearer?

src/util/fpu_lib.c
```diff
@@ -407,52 +407,76 @@ slow_path uint32_t fpu_fclass64(fpu_f64_t d)
 
 slow_path fpu_f32_t fpu_round_f32_internal(fpu_f32_t f, uint32_t mode)
 {
-    uint32_t u = fpu_bit_f32_to_u32(f);
-    uint32_t s = u & FPU_LIB_FP32_SIGNEDFP_MASK;
+    // A non-fractional value (integer, +/-0, inf, NaN) is already rounded. Return it
+    // unchanged: the round-to-nearest path below adds +/-0.5, which is inexact once
+    // 0.5 underflows a large value's ULP and would raise a spurious INEXACT -- e.g.
+    // on an out-of-range fcvt-to-int, leaving NX wrongly set alongside NV.
+    if (likely(!fpu_is_fractional32(f))) {
```
purplesyringa, line 414: I'm not sure what the intention of this `if` is: if it's just to avoid raising wrong exceptions, you're already preventing that with the `fpu_get_exception` logic below. I don't necessarily think adding a fast path here is wrong, but it seems out of scope unless I'm missing something.

src/util/fpu_lib.c
```diff
@@ -407,52 +407,76 @@ slow_path uint32_t fpu_fclass64(fpu_f64_t d)
...
         case FPU_LIB_ROUND_UP:
-            if (!s && fpu_is_fractional32(f)) {
-                return fpu_add32(f, fpu_bit_u32_to_f32(0x3F800000U));
+            if (!s) {
+                r = fpu_add32(f, fpu_bit_u32_to_f32(0x3F800000U));
             }
             break;
     }
-    return f;
+    fpu_set_exceptions(exc);
```
purplesyringa, line 445: Have you benchmarked this? If this is a measurable slow-down, here's an alternative to the "remove fractional `if`" advice: maybe look at `fpu_exponent32(f)` to detect which group it belongs to and choose the path appropriately, so that exceptions are never raised:

- `f` is large enough that it's already necessarily whole and thus doesn't need rounding.
- `f` is in a range where adding `0.5` is always exact.
- `f` is small enough that its rounded value can be clearly determined to be `0` (or maybe `1`, depending on the rounding mode; I trust you can check this case-by-case yourself).

purplesyringa: Hmm, nevermind, I don't think there's actually necessarily a range with exact addition like that. e.g. adding `0.1` to `1111.10001010101` (in binary) will cause an inexact result even though the exponent is quite small.

If exceptions are slow, maybe this is a good reason to manipulate the mantissa manually? We have some similar code in `fpu_is_fractional32`, and it doesn't seem too complex as to be clearly worse than setting exceptions.

Alternatively, maybe always rounding to zero and making the rounding decision based on the fractional part is a good idea.

src/util/fpu_lib.c
```diff
@@ -407,52 +407,76 @@ slow_path uint32_t fpu_fclass64(fpu_f64_t d)
...
+    const uint32_t s   = fpu_bit_f32_to_u32(f) & FPU_LIB_FP32_SIGNEDFP_MASK;
+    fpu_f32_t r = f;
     if (unlikely(mode > FPU_LIB_ROUND_UP)) {
         mode = fpu_get_rounding_mode();
     }
     switch (mode) {
         case FPU_LIB_ROUND_NE:
         case FPU_LIB_ROUND_MM:
-            return fpu_add32(f, fpu_bit_u32_to_f32(0x3F000000U | s));
+            r = fpu_add32(f, fpu_bit_u32_to_f32(0x3F000000U | s));
```
purplesyringa, line 432: Now that I think about it, I don't think this logic actually ever worked? If this addition is inexact, then it absolutely can round in the wrong direction. Consider `f = prev(0.5)`, where `prev` denotes the float just before `0.5`. Then:

- `f = 0.0111111111111111111111111_2`
- `0.5 = 0.1_2`
- `f + 0.5 = 0.1111111111111111111111111_2`, which is inexact, so if the rounding mode in the environment is round-to-nearest, this rounds to `1`. But `round(f)` should've been `0`, not `1`.

*Maybe* you can fix this by temporarily changing the active rounding mode to flooring, but at this point maybe inspecting the fractional part by hand is more efficient.

src/cpu/riscv_fpu.c
```diff
@@ -406,4 +561,32 @@ slow_path func_opt_size void riscv_emulate_f_opc_op(rvvm_hart_t* vm, const uint3
...
 }
 
+/*
+ * RISC-V selects the rounding mode either dynamically (the frm CSR, when the
+ * instruction's rm field is DYN) or statically (the rm field itself). The host
+ * FPU tracks frm, so a static rm field that differs from frm must be applied
+ * around the op. RMM has no host mode at all and is synthesized in RNE by the
+ * fixups above (riscv_rmm_apply / riscv_rmm_div_apply), so its sub-ops must run
+ * in RNE. funct3 == rm only carries a rounding mode on rounding-capable ops, so
+ * this never misfires on fsgnj/fcmp/fclass/fmv.
```
purplesyringa, line 571: I'm not sure what this means? e.g. `fsgnj.s` has 0 in bits 12-14, so this code will read `rm = 0` and temporarily switch the rounding mode to whatever 0 represents, unless I'm missing something. Shouldn't the mode change be gated to only run on specific instructions?

src/cpu/riscv_fpu.c
```diff
@@ -54,46 +54,182 @@ static const uint32_t riscv_fli_table[32] = {
...
+ * (non-finite results also have a non-zero exponent field and return unchanged).
+ *
+ * For a subnormal result the exact residual rho = a - n*b is representable, so
+ * fma(-n, b, a) recovers it exactly. The true quotient is n + rho/b, so it is the
+ * exact midpoint (a tie, rounded away) iff 2*rho == gap*b, where gap = away - n
+ * (= +/- the subnormal step). |rho| <= |b|*2^-1075, so every value below stays in
+ * range and exact. Flag-isolated, as the residual machinery can raise spurious
+ * exceptions.
+ */
+static forceinline fpu_f32_t riscv_rmm_div_apply_f32(fpu_f32_t n, fpu_f32_t a, fpu_f32_t b)
```
purplesyringa, line 195: nit: why is this named `*_apply_*` when the corresponding fixup functions for addition and multiplication don't have `_apply` in the middle?

src/cpu/riscv_fpu.h
```diff
@@ -126,6 +168,175 @@ static forceinline void riscv_emulate_f_opc_store(rvvm_hart_t* vm, const uint32_
...
+    fpu_f64_t r;
+    if (unlikely(eff == 0x04)) {
+        // RMM: compute in RNE, then apply the roundTiesToAway fixup.
+        const uint32_t host = fpu_get_rounding_mode();
+        fpu_set_rounding_mode(FPU_LIB_ROUND_NE);
+        r = riscv_rmm_fma_apply_f64(fpu_fma64(a, b, c), a, b, c);
+        fpu_set_rounding_mode(host);
+    } else if (rm != 0x07 && eff != frm) {
+        // Static host-native mode that differs from frm.
+        const uint32_t host = fpu_get_rounding_mode();
```
purplesyringa, line 329: This feels like a repetition of the logic in `riscv_emulate_f_opc_op`. Can the two be merged together?

purplesyringa:
LGTM re: `sqrt(-0.0)`, `fnmadd` fix, dropping `func_opt_size`. NACK re: `INEXACT` in rounding, honoring `rm`. Will review the FMA UF logic tomorrow. I feel like RMM is a little too complex to review and should be a separate patch, there's plenty of complexity here without it.

src/cpu/riscv_fpu.h
```diff
@@ -134,32 +134,92 @@ static forceinline void riscv_emulate_f_opc_store(rvvm_hart_t* vm, const uint32_
  * from RNE only on an exact halfway tie, and an exact FMA tie is vanishingly rare
  * — the remaining ties-away gap is shared with the rest of the FP path.)
  */
+// Underflow flag, RISC-V semantics: UF iff the result is tiny *after rounding* and
+// inexact. A subnormal result is unambiguously tiny; a normal result above the
+// smallest normal is not. The only hard case is a result of exactly the smallest
+// normal: some hosts (e.g. aarch64) detect tininess *before* rounding and raise a
```
purplesyringa, line 140: > some hosts (e.g. aarch64) detect tininess *before* rounding

Is this an issue for any other arithmetic operations?

src/cpu/riscv_fpu.h
```diff
@@ -134,32 +134,92 @@ static forceinline void riscv_emulate_f_opc_store(rvvm_hart_t* vm, const uint32_
...
+static forceinline void riscv_fma_fixup_uf32(fpu_f32_t r, fpu_f32_t a, fpu_f32_t b, fpu_f32_t c, uint32_t eff)
+{
+    if (likely((fpu_bit_f32_to_u32(r) & 0x7FFFFFFFU) != 0x00800000U)) {
+        return;  // subnormal UF is genuine; a larger normal result has none
+    }
+    const uint32_t exc = fpu_get_exceptions();
+    if (!(exc & FPU_LIB_FLAG_NX)) {
+        return;  // an exact smallest-normal result never underflows
+    }
+    const fpu_f64_t e = fpu_add64(fpu_mul64(fpu_fcvt_f32_to_f64(a), fpu_fcvt_f32_to_f64(b)),
```
purplesyringa, line 157: If this is exact (which I think it always is, maybe modulo `NaN` propagation due to conversions? but I think it becomes canonical in this case and it's still fine), why does `fpu_fma32` uses more complex math? I think it makes sense to update `fpu_fma32` together with this patch.

purplesyringa: I've submitted a couple more comments which highlight issues with this function, but let me provide more general feedback here.

If f64->f32 conversion sets flags correctly, then this line should be used to implement `fma_f32`, regardless of whether the native architecture supports FMA, so that we always get the correct flags without special handling.

If the conversion doesn't always set flags correctly, then it should be updated to do so, and I think it's still better to use this line.

src/cpu/riscv_fpu.h
```diff
@@ -134,32 +134,92 @@ static forceinline void riscv_emulate_f_opc_store(rvvm_hart_t* vm, const uint32_
...
+}
+
+// f64 has no wider type, so recompute the after-rounding tininess by scaling: a
+// smallest-normal f64 result bounds |a|,|b| (a tiny product forces small operands),
+// so a*2^200 cannot overflow, and fma(a*2^200, b, c*2^200) = (a*b+c)*2^200 lands in
+// the normal range where the fused rounding is unbounded-exponent. The result is
+// tiny iff that scaled magnitude is below the scaled smallest normal (2^-822).
+static forceinline void riscv_fma_fixup_uf64(fpu_f64_t r, fpu_f64_t a, fpu_f64_t b, fpu_f64_t c, uint32_t eff)
+{
+    if (likely((fpu_bit_f64_to_u64(r) & 0x7FFFFFFFFFFFFFFFULL) != 0x0010000000000000ULL)) {
```
purplesyringa, line 175: Please add a comment saying this is `2^-1022`.

src/cpu/riscv_fpu.h
```diff
@@ -134,32 +134,92 @@ static forceinline void riscv_emulate_f_opc_store(rvvm_hart_t* vm, const uint32_
...
+        return;  // subnormal UF is genuine; a larger normal result has none
+    }
+    const uint32_t exc = fpu_get_exceptions();
+    if (!(exc & FPU_LIB_FLAG_NX)) {
```
purplesyringa, line 179: I need a recap on how exceptions work. Are they ever reset implicitly? This condition tries to check whether NX was set by the previous fma operation, but I think in reality it also checks whether it has previously ever been set and not reset.

src/cpu/riscv_fpu.h
```diff
@@ -134,32 +134,92 @@ static forceinline void riscv_emulate_f_opc_store(rvvm_hart_t* vm, const uint32_
...
+        return;  // an exact smallest-normal result never underflows
+    }
+    const fpu_f64_t S = fpu_bit_u64_to_f64(0x4C70000000000000ULL);  // 2^200
+    const uint32_t host = fpu_get_rounding_mode();
+    fpu_set_rounding_mode(eff == 0x04 ? FPU_LIB_ROUND_NE : eff);
+    const fpu_f64_t rs = fpu_fma64(fpu_mul64(a, S), b, fpu_mul64(c, S));  // (a*b+c) * 2^200
+    fpu_set_rounding_mode(host);
+    const bool tiny = (fpu_bit_f64_to_u64(rs) & 0x7FFFFFFFFFFFFFFFULL) < 0x0C90000000000000ULL;  // |rs| < 2^-822
+    fpu_set_exceptions(tiny ? (exc | FPU_LIB_FLAG_UF) : (exc & ~FPU_LIB_FLAG_UF));
```
purplesyringa, line 188: same concern re: exceptions, this unsets UF which I think may have been set by a different operation than FMA.

src/cpu/riscv_fpu.h
purplesyringa: Honestly I think this condition can be just removed? It seems to be just a fast path, but accessing exceptions is slow anyway.

src/cpu/riscv_fpu.h
```diff
@@ -134,32 +134,92 @@ static forceinline void riscv_emulate_f_opc_store(rvvm_hart_t* vm, const uint32_
...
+static forceinline void riscv_fma_fixup_uf64(fpu_f64_t r, fpu_f64_t a, fpu_f64_t b, fpu_f64_t c, uint32_t eff)
+{
+    if (likely((fpu_bit_f64_to_u64(r) & 0x7FFFFFFFFFFFFFFFULL) != 0x0010000000000000ULL)) {
+        return;  // subnormal UF is genuine; a larger normal result has none
+    }
+    const uint32_t exc = fpu_get_exceptions();
+    if (!(exc & FPU_LIB_FLAG_NX)) {
+        return;  // an exact smallest-normal result never underflows
+    }
+    const fpu_f64_t S = fpu_bit_u64_to_f64(0x4C70000000000000ULL);  // 2^200
```
purplesyringa, line 182: `2^200` feels a bit arbitrary, it took me a while to figure out where it came from. Maybe change this to `2^52`?

src/cpu/riscv_fpu.h
```diff
@@ -134,32 +134,92 @@ static forceinline void riscv_emulate_f_opc_store(rvvm_hart_t* vm, const uint32_
...
+    const uint32_t host = fpu_get_rounding_mode();
+    fpu_set_rounding_mode(eff == 0x04 ? FPU_LIB_ROUND_NE : eff);
```
purplesyringa, line 184: Why not just call this before resetting the rounding mode back to host in `riscv_fma_round_f64`?

src/cpu/riscv_fpu.h
```diff
@@ -134,32 +134,92 @@ static forceinline void riscv_emulate_f_opc_store(rvvm_hart_t* vm, const uint32_
...
+    const fpu_f32_t rn = fpu_fcvt_f64_to_f32(scaled);
+    fpu_set_rounding_mode(host);
+    const bool tiny = (fpu_bit_f32_to_u32(rn) & 0x7FFFFFFFU) < 0x3F800000U;  // |rn| < 1.0
+    fpu_set_exceptions(tiny ? (exc | FPU_LIB_FLAG_UF) : (exc & ~FPU_LIB_FLAG_UF));
+}
+
+// f64 has no wider type, so recompute the after-rounding tininess by scaling: a
+// smallest-normal f64 result bounds |a|,|b| (a tiny product forces small operands),
+// so a*2^200 cannot overflow, and fma(a*2^200, b, c*2^200) = (a*b+c)*2^200 lands in
```
purplesyringa, line 170: Yeah, but we don't have a tiny product, we have a tiny FMA. "`a*b+c` is tiny" doesn't imply "`a*b` is tiny". You can probably make an argument from precision, but I don't quite see why this works.

purplesyringa: We know that `a*b+c` rounds to `2^-1022` exactly. We don't know what the rounding mode is, but we know that at least one of `a*b` and `c` must include a set bit at `2^-1022` or below -- otherwise, the sum will either produce a greater value or `0`.

If the set bit is in `c`, the maximum value of `|c|` is reached when it's the lowest bits of the mantissa, e.g. `c = 2^-1022 * (2^52 + (2^52 - 1))`, providing `|c| <= 2^-969 - 2^-1022`. By triangle inequality this also gives `|a*b| <= 2^-969`.

If the set bit is in `a*b`, things are a little trickier: since this is FMA, `a*b` is computed exactly and thus has more than 52 bits of precision. Estimating it as 104 bits gives us `|a*b| <= 2^-917 - 2^-1022` and thus `|c| <= 2^-917`.

Taking `|a*b|, |c| <= 2^-917` already demonstrates that computing `c * 2^200` is safe. As long as `b` is non-zero, `|b| >= 2^-1074`, hence `|a| <= 2^157` and computing `a * 2^200` is also safe.

If `b` is zero, `a` is unlimited. Such an FMA is always exact, so the `NX` check passes, but if we remove it (see the relevant comment), it still works out: if `a * 2^200` overflows to `Infinity`, we get `rs = NaN`, and `tiny` ends up being reset because `+NaN` is greater than any finite number.

A similar explanation needs to be included in the code.

purplesyringa:
Reviewed FMA UF handling.

LekKit:
Stupidly, github silently merged #239 together with #240 without asking

PR #240, opened 2026-06-21
riscv_fpu: canonicalize NaN results and mal-boxed narrow operands

SolAstrius:
## Summary

Canonicalize NaN on the scalar F/D path, on both the result and the operand side.

## What's fixed

- **Arithmetic and FMA results** — fadd/fsub/fmul/fdiv/fsqrt and the full FMA family (fmadd/fmsub/fnmsub/fnmadd) route their result through the canonicalizing write path, so a NaN result is delivered as the canonical quiet NaN.
- **Mal-boxed narrow operands** — a narrow (f32) operand that is not correctly NaN-boxed in its 64-bit register reads as the canonical NaN per the RISC-V NaN-boxing rule, across arithmetic, conversions, sign-injection, min/max, compares, classification, and the FMA family.

## Validation

Together with the rounding/exception work in #239, riscv-arch-test (ACT4, Spike reference) F/D/I/M reaches `260/260`.

## Notes

Stacked on #239 — its nine commits appear here until it merges, after which only the two canonical-NaN commits remain. No functional change outside `src/cpu/riscv_fpu.{c,h}`.

### DIFF
--- a/src/cpu/riscv_fpu.c
+++ b/src/cpu/riscv_fpu.c
@@ -68,7 +68,8 @@ static const uint32_t riscv_fli_table[32] = {
  * (The previous implementation rounded toward +/-inf unconditionally, which is
  * correct on ties but wrong for every inexact non-tie. See issue #204.)
  *
- * fadd/fsub/fmul use riscv_rmm_apply (below). fsqrt never needs a fixup: a square
+ * fadd/fsub/fmul use riscv_rmm_apply (in riscv_fpu.h, shared with the FMA family).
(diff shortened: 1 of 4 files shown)


purplesyringa:
NaN canonicalization looks fine, though @lekkit says it should probably be behind a flag. I'm not sure how to implement this efficiently, though. Maybe it's fine to just put an `if` in there since it's an interpreter.

Replacing `riscv_view_s` with `riscv_read_s` seems unquestionably correct, I don't think the spec permits any optimizations here.

PR #241, opened 2026-06-22
riscv_fpu: pre-approved scalar FPU conformance fixes

SolAstrius:
## Summary

The four scalar-FPU fixes that were already reviewed and marked **LGTM** in #239 / #240.

## What's here

- **Drop `func_opt_size` on the FP op dispatch.** The scalar FP path is interpreted, not JIT'd, so size-optimizing the dispatch only costs throughput — roughly 2x interpreter FP at zero size cost elsewhere.
- **`sqrt(-0.0)` returns `-0.0` without raising invalid.** IEEE 754 `squareRoot(-0)` is `-0` with no exception; only a negative non-zero (or `-inf`) operand is invalid. The `-0` carve-out is gated on a new `FPU_LIB_FPxx_NEGATIVE_ZERO` constant so the intent reads clearly (per review feedback).
- **fnmadd negates the operands, not the result.** `fnmadd` computes `-(rs1*rs2) - rs3` as a single fused op; negating `fma(rs1, rs2, rs3)` instead rounds `rs1*rs2 + rs3` and then flips the sign, which under directed rounding modes can land on the wrong side. Negate `rs1` and `rs3` so the one rounding sees the correctly-signed value.
- **Mal-boxed narrow operands read as the canonical NaN.** Per the RISC-V NaN-boxing rule, a narrow (f32) operand that isn't correctly NaN-boxed in its 64-bit register must read as the canonical NaN. Value reads route through `riscv_read_s` across arithmetic, conversions, sign-injection, min/max, compares, classification, and the FMA family. `fmv.x.w` keeps the raw `riscv_view_s` read (it moves the register bits verbatim).

### DIFF
--- a/src/cpu/riscv_fpu.c
+++ b/src/cpu/riscv_fpu.c
@@ -93,7 +93,7 @@ static void riscv_prepare_rmm(rvvm_hart_t* vm, const uint32_t insn, const size_t
     }
 }
 
-slow_path func_opt_size void riscv_emulate_f_opc_op(rvvm_hart_t* vm, const uint32_t insn)
+slow_path void riscv_emulate_f_opc_op(rvvm_hart_t* vm, const uint32_t insn)
 {
     const size_t   rds = bit_ext_u32(insn, 7, 5);
     const uint32_t rm  = bit_ext_u32(insn, 12, 3);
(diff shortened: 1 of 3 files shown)


SolAstrius:
@purplesyringa

purplesyringa (approved):

PR #242, opened 2026-06-22
riscv_fpu: RMM (roundTiesToAway) for the scalar OP-FP ops

SolAstrius:
## Summary

Fix RMM (round to nearest, ties to max magnitude — IEEE 754 `roundTiesToAway`) for the scalar arithmetic ops, addressing #204. Standalone on `staging`; no dependency on the other in-flight FPU work.

The host FPU has no RMM mode, so under `frm == RMM` the host is left in round-to-nearest-even. RNE and roundTiesToAway differ only on an exact halfway tie. The previous `riscv_prepare_rmm` rounded toward ±inf unconditionally — correct on ties but wrong for every inexact non-tie. This replaces it with: compute in RNE, recover the **exact** rounding error via the library's error-free transforms (TwoSum / TwoProduct), and step one ULP outward only on a true half-ULP tie.

## What's covered

- **fadd / fsub / fmul / fdiv**, both `.s` and `.d`.
- **fmul near the underflow boundary**, where Dekker's product error itself underflows — handled with an exact widened (f32) / scaled-residual (f64) tie test.
- **subnormal fdiv** quotients, the one division case that can land on an exact tie.
- fsqrt needs no fixup (a square root is never an exact halfway case).
- Flag-isolated throughout: the error-free transforms do raw host arithmetic whose intermediate steps can raise spurious NV/OF, so each wrapper snapshots and restores the exception flags; the genuine flags come from the base op.

## Scope / draft notes

Deliberately minimal and independent, so it can be reviewed on its own (per review feedback that RMM is too complex to evaluate bundled with the rest):

- **Dynamic RMM only** (`frm == RMM`, `rm == DYN`). The static `,rmm` instruction suffix needs host-rounding-mode forcing and lands with the rounding-mode rework.
- **OP-FP only — no FMA.** The FMA-family ties-away fixup depends on the FMA rounding path; under RMM the FMA ops still round ties-to-even here, to be completed with the FMA work.

Draft until the companion FPU PRs settle (it lightly overlaps the fadd/fsub/fmul/fdiv read sites with the canonicalization/mal-box PR, so ordering may want a trivial rebase).

Touches `src/cpu/riscv_fpu.c` only.

### DIFF
--- a/src/cpu/riscv_fpu.c
+++ b/src/cpu/riscv_fpu.c
@@ -138,14 +138,62 @@ static forceinline fpu_f64_t riscv_rmm_add_f64(fpu_f64_t n, fpu_f64_t a, fpu_f64
     return r;
 }
 
-// roundTiesToAway fixup for fmul. Flag-isolated.
+/*
+ * Dekker's product error (fpu_mul_error) loses bits once the product error drops
+ * below the smallest subnormal -- not only for a subnormal result, but throughout
+ * the smallest normal binades (the half-ULP ~2^(e-53) underflows for e <= -969).
+ * Those results need an exact tie test.
+ *
+ * For f32 the exact product widens losslessly into f64 (48 <= 53 bits) and the
+ * away-side midpoint is exact there (<= 25 bits, down to 2^-150), so the op is a
+ * tie iff a*b == m. This is exact for every f32 result, so fmul.s always uses it.
+ *
+ * For f64 there is no wider type, so scale both operands by 2^512 to lift the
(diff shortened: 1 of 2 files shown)


SolAstrius:
@purplesyringa addressed the #239 comments that apply here: the RMM fixups now live in fpu_lib next to the error-free transforms, with a header note that RMM is riscv-specific (as agreed), and the div_apply naming inconsistency is gone. Dynamic-frm RMM harness passes 21826/21826; static rm handling stays a separate patch. Keeping it draft so it doesn't get merged by accident again — review when you have time.

PR #246, opened 2026-07-17
riscv_fpu: static/FMA rounding modes, exact integral rounding, FMA underflow tininess

SolAstrius:
## Summary

Reworked resubmission of the remaining #239 commits, addressing @purplesyringa's review. Four commits, each standalone on `staging`; no dependency on #242 (RMM), and the two deliberately overlap nowhere except the shared `riscv_fpu_host_rm` helper.

### 1. `fpu_lib: round to integral values on the encoding, not via biasing adds`

Replaces the NACK'd INEXACT patch. The review showed the old `f + 0.5` trick was broken beyond the flag leak ([this comment](https://github.com/LekKit/RVVM/pull/239#discussion_r3449015750)): the biasing add itself rounds, so `rne(prev(0.5))` returned 1 and `rne(2.5)` returned 3 (ties-away, not ties-to-even). `fpu_round_fXX_internal` now decides on the fraction bits of the encoding: no host FP arithmetic, so no spurious flags, no double rounding, and no dependence on a host rounding mode (relevant for the future `USE_SOFT_FENV` story). RMM is a first-class case — only DYN falls back to the tracked mode, so `fcvt`/`fround` with a static `rmm` field round correctly.

Verified bit-exact against host `rint()`/`round()` in all four modes over the full fractional exponent range of both widths (2.0e9 cases, zero mismatches).

### 2. `riscv_fpu: honor the static rounding-mode field`

Reworked per review: the mode change is now gated per-instruction (`riscv_f_op_is_rounding`) — funct3 is an rm field only on rounding-capable ops, so `fsgnj`/`fcmp`/`fclass`/`fmv` never trigger it. Ops that consume `rm` as an argument (fcvt-to-int, fround) need no wrap. A static `rmm` field runs in RNE for now (differs only on exact ties; wired to the exact fixups after #242 lands). The dynamic-RMM preparation is gated to `rm == DYN` so a static field is no longer clobbered by it.

### 3. `riscv_fpu: honor the rounding mode in the FMA ops`

The fnmadd operand-negation part already landed via #241; this is the rm plumbing, now sharing `riscv_fpu_host_rm` with the OP-FP dispatch instead of duplicating the effective-mode logic (review ask).

### 4. `fpu_lib: set the FMA underflow flag by IEEE after-rounding tininess`

Reworked per review, and moved into `fpu_fma32/64` proper; `fpu_fma32`'s own rounding is unchanged (see the review thread for why the direct downcast mis-rounds):

- the NX fast-path gate is gone (it conflated "set by this op" with "ever set");
- the verdict merges over a flag snapshot from before the op, so a sticky UF from earlier ops survives;
- the fixup runs while the effective mode is still set (it lives inside the op now);
- f64 scales by 2^52 instead of 2^200, with the operand-bound proof from the review included as a comment; the `b == 0` overflow-to-NaN case reads correctly as not-tiny;
- f32 decides tininess by comparing the widened 53-bit sum against the mode's threshold in f64, with the exact TwoSum error breaking the tie when the sum lands exactly on the to-nearest threshold.

On aarch64 (before-rounding tininess), `fma(2^-62*(1-2^-23), 2^-64*(1+2^-23), 0)` returned the correct 2^-126 but flagged NX|UF instead of NX; targeted tests cover both boundary directions, an exact boundary, a genuinely-tiny-at-boundary case, and sticky-UF preservation — all pass with the fix, three fail without it.

## Conformance (riscv-arch-test ACT4, Spike reference)

| build | F | D | I | M | total |
|---|---|---|---|---|---|
| staging | 13/82 | 24/114 | 51/51 | 13/13 | 101/260 |
| this PR | 21/82 | 51/114 | 51/51 | 13/13 | **136/260** |

No regressions. The remaining F/D failures are, by sampled diagnosis, exactly the two intentionally-excluded categories:

- **RMM vectors** — the old `riscv_prepare_rmm` directed hack (e.g. it reads a NaN-boxed f32 through the f64 `default:` path and rounds fsqrt the wrong way). Deleted and replaced by #242.
- **NaN result canonicalization** — arithmetic results emit host NaN payloads where the spec wants the canonical NaN (`fadd.s` → `0x7fd825fe` vs `0x7fc00000`). That's the separate flag-gated patch we agreed on; needs a decision on compile-time vs runtime flag first.

A local combined build (this PR + #242 + the canonicalization patch) reaches **244/260**, and every one of its 16 residual failures is an FMA test failing on an `frm=rmm` tie sub-case — that's the one #239 commit not yet resubmitted anywhere ("round the FMA family to nearest, ties-away under RMM"); it needs #242's fixup machinery, so it goes on top of #242 once that merges, closing the gap to the 260/260 the old chain reported.

cc @purplesyringa — draft until reviewed, to keep GitHub from merging things on its own again.

### DIFF
--- a/src/cpu/riscv_fpu.c
+++ b/src/cpu/riscv_fpu.c
@@ -81,8 +81,9 @@ static void riscv_prepare_rmm(rvvm_hart_t* vm, const uint32_t insn, const size_t
             neg = fpu_signbit64(riscv_view_d(vm, rs1)) != fpu_signbit64(riscv_view_d(vm, rs2));
             break;
         default:
-            neg = fpu_signbit64(riscv_view_d(vm, rs1));
-            break;
+            // Only add/sub/mul/div need the directed synthesis: sqrt has no exact
+            // ties, and ops taking rm as an argument handle RMM natively
+            return;
     }
 
     // Round to positive/negative infinity based on the result sign
--- a/src/cpu/riscv_fpu.c
+++ b/src/cpu/riscv_fpu.c
@@ -94,8 +94,10 @@ static void riscv_prepare_rmm(rvvm_hart_t* vm, const uint32_t insn, const size_t
 }
 
 // funct3 is an rm field only on the rounding-capable OP-FP ops; on
-// fsgnj/fmin/fmax/fcmp/fclass/fmv it encodes the operation itself
-static forceinline bool riscv_f_op_is_rounding(const uint32_t insn)
+// fsgnj/fmin/fmax/fcmp/fclass/fmv it encodes the operation itself. "Implicitly"
+// rounding: ops that take rm as an argument (fcvt to integer, fround) consume
+// the field themselves and are not listed here.
+static forceinline bool riscv_f_op_is_implicitly_rounding(const uint32_t insn)
 {
     switch (insn & 0xFE000000UL) {
         case 0x00000000UL: // fadd.s
@@ -126,8 +128,9 @@ static slow_path void riscv_emulate_f_opc_op_impl(rvvm_hart_t* vm, const uint32_
 
     if (likely(riscv_fpu_is_enabled(vm))) {
 
-        if (unlikely(vm->csr.fcsr >> 5 == 0x04) && rm == 0x07) {
-            // Handle dynamic RMM rounding; a static rm field overrides frm
+        if (unlikely(((rm == 0x07) ? vm->csr.fcsr >> 5 : rm) == 0x04)) {
+            // Handle RMM rounding in the effective mode: a static rmm field
+            // behaves exactly like frm == RMM
             riscv_prepare_rmm(vm, insn, rs1, rs2);
         }
 
@@ -433,19 +436,15 @@ static slow_path void riscv_emulate_f_opc_op_impl(rvvm_hart_t* vm, const uint32_
 slow_path void riscv_emulate_f_opc_op(rvvm_hart_t* vm, const uint32_t insn)
 {
     const uint32_t rm = bit_ext_u32(insn, 12, 3);
-    // A static rm field on a rounding-capable op overrides the frm-tracked host
-    // mode around the op; ops taking rm as an argument consume the field directly
-    if (unlikely(rm != 0x07) && riscv_fpu_rm_is_valid(rm) && riscv_f_op_is_rounding(insn)) {
-        const uint32_t prev = fpu_get_rounding_mode();
-        const uint32_t need = riscv_fpu_host_rm(rm);
-        if (need != prev) {
-            fpu_set_rounding_mode(need);
-            riscv_emulate_f_opc_op_impl(vm, insn);
-            fpu_set_rounding_mode(prev);
-            return;
-        }
+    // A static rm field on an implicitly rounding op overrides the frm-tracked
+    // host mode around the op
+    if (unlikely(rm != 0x07) && riscv_fpu_rm_is_valid(rm) && riscv_f_op_is_implicitly_rounding(insn)) {
+        const uint32_t prev_rm = riscv_fpu_static_rm_enter(rm);
+        riscv_emulate_f_opc_op_impl(vm, insn);
+        riscv_fpu_static_rm_leave(prev_rm);
+    } else {
+        riscv_emulate_f_opc_op_impl(vm, insn);
     }
-    riscv_emulate_f_opc_op_impl(vm, insn);
 }
 
 #endif
--- a/src/cpu/riscv_fpu.c
+++ b/src/cpu/riscv_fpu.c
@@ -93,7 +93,31 @@ static void riscv_prepare_rmm(rvvm_hart_t* vm, const uint32_t insn, const size_t
     }
 }
 
-slow_path void riscv_emulate_f_opc_op(rvvm_hart_t* vm, const uint32_t insn)
+// funct3 is an rm field only on the rounding-capable OP-FP ops; on
+// fsgnj/fmin/fmax/fcmp/fclass/fmv it encodes the operation itself
+static forceinline bool riscv_f_op_is_rounding(const uint32_t insn)
+{
+    switch (insn & 0xFE000000UL) {
+        case 0x00000000UL: // fadd.s
+        case 0x02000000UL: // fadd.d
+        case 0x08000000UL: // fsub.s
+        case 0x0A000000UL: // fsub.d
+        case 0x10000000UL: // fmul.s
+        case 0x12000000UL: // fmul.d
+        case 0x18000000UL: // fdiv.s
+        case 0x1A000000UL: // fdiv.d
+        case 0x58000000UL: // fsqrt.s
+        case 0x5A000000UL: // fsqrt.d
+        case 0x40000000UL: // fcvt.s.d, fround.s (Zfa)
+        case 0x42000000UL: // fcvt.d.s, fround.d (Zfa)
+        case 0xD0000000UL: // fcvt.s.w[u]/l[u]
+        case 0xD2000000UL: // fcvt.d.w[u]/l[u]
+            return true;
+    }
+    return false;
+}
+
+static slow_path void riscv_emulate_f_opc_op_impl(rvvm_hart_t* vm, const uint32_t insn)
 {
     const size_t   rds = bit_ext_u32(insn, 7, 5);
     const uint32_t rm  = bit_ext_u32(insn, 12, 3);
@@ -102,8 +126,8 @@ slow_path void riscv_emulate_f_opc_op(rvvm_hart_t* vm, const uint32_t insn)
 
     if (likely(riscv_fpu_is_enabled(vm))) {
 
-        if (unlikely(vm->csr.fcsr >> 5 == 0x04)) {
-            // Handle RMM rounding
+        if (unlikely(vm->csr.fcsr >> 5 == 0x04) && rm == 0x07) {
+            // Handle dynamic RMM rounding; a static rm field overrides frm
             riscv_prepare_rmm(vm, insn, rs1, rs2);
         }
 
@@ -404,6 +428,24 @@ slow_path void riscv_emulate_f_opc_op(rvvm_hart_t* vm, const uint32_t insn)
     }
 
     riscv_illegal_insn(vm, insn);
+}
+
+slow_path void riscv_emulate_f_opc_op(rvvm_hart_t* vm, const uint32_t insn)
+{
+    const uint32_t rm = bit_ext_u32(insn, 12, 3);
+    // A static rm field on a rounding-capable op overrides the frm-tracked host
+    // mode around the op; ops taking rm as an argument consume the field directly
+    if (unlikely(rm != 0x07) && riscv_fpu_rm_is_valid(rm) && riscv_f_op_is_rounding(insn)) {
+        const uint32_t prev = fpu_get_rounding_mode();
+        const uint32_t need = riscv_fpu_host_rm(rm);
+        if (need != prev) {
+            fpu_set_rounding_mode(need);
+            riscv_emulate_f_opc_op_impl(vm, insn);
+            fpu_set_rounding_mode(prev);
+            return;
+        }
+    }
+    riscv_emulate_f_opc_op_impl(vm, insn);
 }
 
 #endif
--- a/src/cpu/riscv_fpu.h
+++ b/src/cpu/riscv_fpu.h
@@ -41,11 +41,11 @@ static forceinline uint32_t riscv_fpu_host_rm(uint32_t rm)
 static forceinline uint32_t riscv_fpu_static_rm_enter(uint32_t rm)
 {
     if (unlikely(rm != 0x07)) {
+        // Always report a mode to restore: the op itself may change it further
+        // (the RMM preparation), and leave must undo that too
         const uint32_t prev = fpu_get_rounding_mode();
-        if (riscv_fpu_host_rm(rm) != prev) {
-            fpu_set_rounding_mode(riscv_fpu_host_rm(rm));
-            return prev;
-        }
+        fpu_set_rounding_mode(riscv_fpu_host_rm(rm));
+        return prev;
     }
     return 0x07;
 }
--- a/src/cpu/riscv_fpu.h
+++ b/src/cpu/riscv_fpu.h
@@ -33,6 +33,30 @@ static forceinline uint32_t riscv_fpu_host_rm(uint32_t rm)
     return (rm == FPU_LIB_ROUND_MM) ? FPU_LIB_ROUND_NE : rm;
 }
 
+/*
+ * A rounding-capable op runs under its effective mode: the frm-tracked host mode
+ * when rm == DYN, or the static rm field applied around the op. Enter returns the
+ * mode to restore afterwards via leave, or DYN when nothing was changed.
+ */
+static forceinline uint32_t riscv_fpu_static_rm_enter(uint32_t rm)
+{
+    if (unlikely(rm != 0x07)) {
+        const uint32_t prev = fpu_get_rounding_mode();
+        if (riscv_fpu_host_rm(rm) != prev) {
+            fpu_set_rounding_mode(riscv_fpu_host_rm(rm));
+            return prev;
+        }
+    }
+    return 0x07;
+}
+
+static forceinline void riscv_fpu_static_rm_leave(uint32_t prev)
+{
+    if (unlikely(prev != 0x07)) {
+        fpu_set_rounding_mode(prev);
+    }
+}
+
 // Bit-precise register read (raw low 32 bits, no NaN-box check) -- for fmv.x.w
 static forceinline fpu_f32_t riscv_view_s(rvvm_hart_t* vm, size_t reg)
 {
@@ -133,38 +157,6 @@ static forceinline void riscv_emulate_f_opc_store(rvvm_hart_t* vm, const uint32_
     riscv_illegal_insn(vm, insn);
 }
 
-// FMA in the op's rounding mode: a static rm field that differs from the
-// frm-tracked host mode is applied around the op, as in the OP-FP dispatch
-static forceinline fpu_f32_t riscv_fma32(uint32_t rm, fpu_f32_t a, fpu_f32_t b, fpu_f32_t c)
-{
-    if (unlikely(rm != 0x07)) {
-        const uint32_t prev = fpu_get_rounding_mode();
-        const uint32_t need = riscv_fpu_host_rm(rm);
-        if (need != prev) {
-            fpu_set_rounding_mode(need);
-            const fpu_f32_t r = fpu_fma32(a, b, c);
-            fpu_set_rounding_mode(prev);
-            return r;
-        }
-    }
-    return fpu_fma32(a, b, c);
-}
-
-static forceinline fpu_f64_t riscv_fma64(uint32_t rm, fpu_f64_t a, fpu_f64_t b, fpu_f64_t c)
-{
-    if (unlikely(rm != 0x07)) {
-        const uint32_t prev = fpu_get_rounding_mode();
-        const uint32_t need = riscv_fpu_host_rm(rm);
-        if (need != prev) {
-            fpu_set_rounding_mode(need);
-            const fpu_f64_t r = fpu_fma64(a, b, c);
-            fpu_set_rounding_mode(prev);
-            return r;
-        }
-    }
-    return fpu_fma64(a, b, c);
-}
-
 static forceinline void riscv_emulate_f_fmadd(rvvm_hart_t* vm, const uint32_t insn)
 {
     const size_t   rds = bit_ext_u32(insn, 7, 5);
@@ -174,22 +166,25 @@ static forceinline void riscv_emulate_f_fmadd(rvvm_hart_t* vm, const uint32_t in
     const size_t   rs3 = insn >> 27;
 
     if (likely(riscv_fpu_is_enabled(vm) && riscv_fpu_rm_is_valid(rm))) {
+        // A static rm field overrides the frm-tracked host mode, as in the OP-FP dispatch
+        const uint32_t prev_rm = riscv_fpu_static_rm_enter(rm);
         switch (bit_ext_u32(insn, 25, 2)) {
             case 0x0: // fmadd.s
                 riscv_emit_s(vm, rds,
-                             riscv_fma32(rm,                    //
-                                         riscv_read_s(vm, rs1), //
-                                         riscv_read_s(vm, rs2), //
-                                         riscv_read_s(vm, rs3)));
+                             fpu_fma32(riscv_read_s(vm, rs1), //
+                                       riscv_read_s(vm, rs2), //
+                                       riscv_read_s(vm, rs3)));
+                riscv_fpu_static_rm_leave(prev_rm);
                 return;
             case 0x1: // fmadd.d
                 riscv_emit_d(vm, rds,
-                             riscv_fma64(rm,                    //
-                                         riscv_view_d(vm, rs1), //
-                                         riscv_view_d(vm, rs2), //
-                                         riscv_view_d(vm, rs3)));
+                             fpu_fma64(riscv_view_d(vm, rs1), //
+                                       riscv_view_d(vm, rs2), //
+                                       riscv_view_d(vm, rs3)));
+                riscv_fpu_static_rm_leave(prev_rm);
                 return;
         }
+        riscv_fpu_static_rm_leave(prev_rm);
     }
 
     riscv_illegal_insn(vm, insn);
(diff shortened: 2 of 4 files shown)


src/util/fpu_lib.c
```diff
@@ -405,54 +405,109 @@ slow_path uint32_t fpu_fclass64(fpu_f64_t d)
...
+                away = !!s;
+                break;
+            case FPU_LIB_ROUND_UP:
+                away = !s;
+                break;
+        }
+        return fpu_bit_u64_to_f64(s | (away ? 0x3FF0000000000000ULL : 0));
+    }
+    const uint64_t frac = u & (FPU_LIB_FP64_MANTISSA_MASK >> e);
+    const uint64_t half = 0x0008000000000000ULL >> e;
```
purplesyringa: Maybe use `1ULL << 51` here for readability? And same for `step` and the 32-bit case.

src/util/fpu_lib.c
```diff
@@ -405,54 +405,109 @@ slow_path uint32_t fpu_fclass64(fpu_f64_t d)
...
             break;
         case FPU_LIB_ROUND_UP:
-            if (!s && fpu_is_fractional64(d)) {
-                return fpu_add64(d, fpu_bit_u64_to_f64(0x3FF0000000000000ULL));
-            }
+            away = !s && frac;
             break;
     }
-    return d;
+    return fpu_bit_u64_to_f64((u - frac) + (away ? step : 0));
```
purplesyringa, line 519: This works even if the addition overflows into the exponent in situations like `1111'0000 + 0001'0000 = 1'0000'0000`, because that only happens for powers of two, for which incrementing the exponent produces the intended behavior (see e.g. https://blog.m-ou.se/floats/#tricks). Could this be a comment? (probably in the 32-bit path, I mistakenly read the 64-bit code first)

src/util/fpu_lib.c
```diff
@@ -405,54 +405,109 @@ slow_path uint32_t fpu_fclass64(fpu_f64_t d)
...
         mode = fpu_get_rounding_mode();
     }
+    if (e >= 23 || !(u << 1)) {
+        return f; // Already integral: |f| >= 2^23, +/-0, inf, NaN
+    }
+    if (e < 0) {
+        // |f| in (0, 1) rounds to +/-0 or +/-1; the 0.5 tie goes to even == 0
+        switch (mode) {
+            case FPU_LIB_ROUND_NE:
+                away = (e == -1) && (u & FPU_LIB_FP32_MANTISSA_MASK);
```
purplesyringa, line 431: Confused about this `&&`: shouldn't `1.0 * 2^-1 = 1/2` round to `1` despite having a full-zero mantissa?

src/util/fpu_lib.c
```diff
@@ -405,54 +405,109 @@ slow_path uint32_t fpu_fclass64(fpu_f64_t d)
...
+        }
+        return fpu_bit_u32_to_f32(s | (away ? 0x3F800000U : 0));
+    }
+    // |f| in [1, 2^23): split the mantissa into integer part and fraction bits
+    const uint32_t frac = u & (FPU_LIB_FP32_MANTISSA_MASK >> e);
+    const uint32_t half = 0x00400000U >> e;
+    const uint32_t step = 0x00800000U >> e; // 1.0 at this exponent; carries into a new binade
     switch (mode) {
         case FPU_LIB_ROUND_NE:
+            away = frac > half || (frac == half && (u & step));
```
purplesyringa, line 454: If `e = 0`, the `u & step` condition checks the lowest bit of the (biased) exponent. `e = 0` corresponds to an odd stored exponent, so this evaluates to `true`, and `1.5` correctly gets rounded up. Could you add a comment covering this edge case?

src/util/fpu_lib.c
```diff
@@ -405,54 +405,109 @@ slow_path uint32_t fpu_fclass64(fpu_f64_t d)
...
+                away = (e == -1);
+                break;
+            case FPU_LIB_ROUND_DN:
+                away = !!s;
+                break;
+            case FPU_LIB_ROUND_UP:
+                away = !s;
+                break;
+        }
+        return fpu_bit_u32_to_f32(s | (away ? 0x3F800000U : 0));
```
purplesyringa: `/* 1.0 */`

src/cpu/riscv_fpu.c
```diff
@@ -102,8 +126,8 @@ slow_path void riscv_emulate_f_opc_op(rvvm_hart_t* vm, const uint32_t insn)
 
     if (likely(riscv_fpu_is_enabled(vm))) {
 
-        if (unlikely(vm->csr.fcsr >> 5 == 0x04)) {
-            // Handle RMM rounding
+        if (unlikely(vm->csr.fcsr >> 5 == 0x04) && rm == 0x07) {
```
purplesyringa: Shouldn't this logic also trigger for static RMM? I don't see why the semantics of static RMM would differ from dynamic RMM. The commit mentions #242, so I'm assuming this is temporary, but I don't see why it has to be written this way.

purplesyringa:
I noticed that the first commit (the one about float-to-integer rounding) kept

```c
    if (unlikely(fpu_is_zero64_soft(d))) {
        return 0;
    }
```

in `fpu_round_f64_to_u32` (and similar methods). I'm not sure what role that `if` used to play, but judging from `unlikely`, maybe it's not needed anymore? @LekKit do you remember why that was needed?

src/cpu/riscv_fpu.c
```diff
@@ -93,7 +93,31 @@ static void riscv_prepare_rmm(rvvm_hart_t* vm, const uint32_t insn, const size_t
     }
 }
 
-slow_path void riscv_emulate_f_opc_op(rvvm_hart_t* vm, const uint32_t insn)
+// funct3 is an rm field only on the rounding-capable OP-FP ops; on
+// fsgnj/fmin/fmax/fcmp/fclass/fmv it encodes the operation itself
+static forceinline bool riscv_f_op_is_rounding(const uint32_t insn)
```
purplesyringa: Maybe `riscv_f_op_is_implicitly_rounding` to disambiguate it from ops that handle rounding manually?

purplesyringa:
Just throwing ideas out there: I think fp-to-int roundings can be optimized further. I don't expect you to do this here if you don't want to, I'm mostly just glad it's sound, but you can try this if you have time.

The old FP-based code tried to force correct roundings with formulas like `f + 0.5`, which have edge cases, because you're trying to fit complex rules in one formula. However, you can force the FPU to do the correct rounding with a different method. Let's say `0 <= f < 2^52 - 1` is a floating-point value. Then `f + 2^52` will have an exponent of exactly 52, so the mantissa field is exactly `round(f)`. Since addition uses the rounding mode from fenv, you can effectively control which rounding mode this method implements. [See also](https://purplesyringa.moe/blog/fast-limited-range-conversion-between-ints-and-floats/).

This method introduces a few subtleties. For instance, there are some edge cases around `f ~= 2^52`, e.g. for `f = 2^52 - 0.5`, `f + 2^52` can round up to `2^53`, which has an empty mantissa field, but you can fix this by computing `bitcast(f + 2^52) - bitcast(2^52)` instead of masking bits. Point is, you'll still need some special casing for `e >= 52`, but you don't need a `switch` on the runtime type, you can just use the host mode for everything and avoid manually implementing FP logic.

Of course, RMM is the elephant in the room. That will still require some sort of fixup.

src/cpu/riscv_fpu.h
```diff
@@ -133,6 +133,38 @@ static forceinline void riscv_emulate_f_opc_store(rvvm_hart_t* vm, const uint32_
     riscv_illegal_insn(vm, insn);
 }
 
+// FMA in the op's rounding mode: a static rm field that differs from the
+// frm-tracked host mode is applied around the op, as in the OP-FP dispatch
+static forceinline fpu_f32_t riscv_fma32(uint32_t rm, fpu_f32_t a, fpu_f32_t b, fpu_f32_t c)
```
purplesyringa: The FP-OP path puts this handling around `riscv_emulate_f_opc_op`, not individual operations. Does it sound reasonable to move it out here as well? At least to reduce codegen bloat (since this is `forceinline`), but also just to consolidate the code paths in source, since they're getting somewhat divergent.

Case in point: this path doesn't check `riscv_fpu_rm_is_valid(rm)`, but the one in `riscv_emulate_f_opc_op` does. Which one is correct?

src/util/fpu_lib.h
purplesyringa, line 1073: [My original review](https://github.com/LekKit/RVVM/pull/239/changes/0468b7dd8da7657d72eb3c0dd99795cb11e33b41#r3450160564) mentioned that computing `sum` is already exact in the 32-bit case, so handling `err/res/soft_fenv_...` is unnecessary and you can directly downcast `sum` to `f32` and obtain the correct result. The PR mentions that you fixed `fpu_fma32`, but I don't see that? Did you run into some problems while trying that?

src/util/fpu_lib.h
```diff
@@ -1024,6 +1053,10 @@ static forceinline func_opt_size fpu_f32_t fpu_fma32(fpu_f32_t a, fpu_f32_t b, f
 #endif
 #endif
 
+    if (unlikely((fpu_bit_f32_to_u32(ret) & FPU_LIB_FP32_NOSIGNED_MASK) == FPU_LIB_FP32_MINIMUM_NORM)) {
```
purplesyringa, line 1080: Hopefully you can explain this to me.

If I understand it correctly, RISC-V says UF corresponds to the value after rounding for *all* operations. In particular, I'm assuming this includes `fcvt.s.d`, which is implemented via `fpu_fcvt_f32_to_f64`.

Since `f32` denormals easily fit in `f64`, the only function in `fpu_fma32` that can set UF is `fpu_fcvt_f32_to_f64`. So if `fcvt.s.d` sets UF correctly, then `fpu_fcvt_f32_to_f64` should be sufficient here would FMA-specific logic, and this function could be much, much simpler (I think you wouldn't need to handle exceptions at all, actually, but I'm not certain about that). This would certainly remove `fpu_fma32_fixup_uf` at least.

And if `fpu_fcvt_f32_to_f64` is broken, then that's where the fix should live, not here, I think. This also raises the topic of whether any other arithmetic operations need similar adjustments -- e.g. what specifically makes FMA so different from multiplication that only FMA needs special handling?

Could you explain where my logic fails?

src/util/fpu_lib.h
```diff
@@ -1034,22 +1067,61 @@ static forceinline func_opt_size fpu_f32_t fpu_fma32(fpu_f32_t a, fpu_f32_t b, f
...
+/*
+ * f64 counterpart of fpu_fma32_fixup_uf: no wider type exists, so rescale the op
+ * by 2^52 into the normal range, where the single fused rounding is already
+ * unbounded-equivalent, and test the scaled result against 2^-970.
+ *
+ * The scaling is safe: the result rounds to +/-2^-1022, so at least one of a*b, c
+ * has a set bit at 2^-1022 or below (else the sum is either 0 or larger). If it is
+ * in c, |c| <= 2^-969 and by triangle inequality |a*b| <= 2^-968; if it is in the
+ * exact product (up to 106 bits wide), |a*b| <= 2^-916 and |c| <= 2^-915. Either
+ * way |c|*2^52 is tiny and, for b != 0, |b| >= 2^-1074 gives |a| <= 2^159, so
```
purplesyringa:
```suggestion
 * way |c|*2^52 is tiny and, for b != 0, |b| >= 2^-1074 gives |a| <= 2^158, so
```

purplesyringa:
I think this about concludes my review for now. The logic I didn't comment on looks good.

LekKit:
> I noticed that the first commit (the one about float-to-integer rounding) kept
> 
> ```c
>     if (unlikely(fpu_is_zero64_soft(d))) {
>         return 0;
>     }
> ```
> 
> in `fpu_round_f64_to_u32` (and similar methods). I'm not sure what role that `if` used to play, but judging from `unlikely`, maybe it's not needed anymore? @LekKit do you remember why that was needed?

AFAIK the original rounding routine was written in a way that did not account for literal zero (which is already rounded anyways). Now this seems redundant, but i would instead verify if it actually produces expected result.

src/util/fpu_lib.c
purplesyringa: Makes sense, thanks.

purplesyringa:
LGTM on everything except the `fpu_fma32` UF handling, I still don't understand why we have some complicated logic here.

Say, consider the example in the commit message. We have an exact result of `2^-126 - 2^-151 = 2^-127 * (1 + (1 - 2^-24))`. Round-to-nearest with an unbounded exponent produces `2^-127 * (1 + (1 - 2^23))`, which is below the smallest denormal, and bounded rounding produces `2^-126`, which is not tiny, so yes, setting `tiny` from the value already rounded to `f32` is incorrect, as the commit description correctly mentions.

But we never used `sum` rounded to `f32` for calculating `tiny`, we first multiplied it by `2^126`, so there was never a situation where the exponent in this rounding would get bounded. And even if it did, we can just compare `sum` to `2^-126` while staying in `f64` -- in fact, that should be faster and cleaner.

So I don't see why you're piling on more logic instead of removing it. See also: the two unresolved review comments on this topic, in particular I'm wondering if other operations, like addition or f64-to-f32 downcasting, should also get special handling for UF, and if so, if 32-bit FMA UF handling would look simpler if it was based on that logic.

SolAstrius:
In b3a3f3a I have introduced a regression; fixing it.

```
== b17bfc5 (fpu_lib: set the FMA underflow flag by IEEE a)
  F-fcvt.l.s-00 PASS
  F-fcvt.s.w-00 PASS
  F-fcvt.wu.s-00 PASS
== 1587da8 (fpu_lib: clarify the integral-rounding edge c)
  F-fcvt.l.s-00 PASS
  F-fcvt.s.w-00 PASS
  F-fcvt.wu.s-00 PASS
== b3a3f3a (riscv_fpu: share static-rm handling between t)
  F-fcvt.l.s-00 FAIL
  F-fcvt.s.w-00 FAIL
  F-fcvt.wu.s-00 FAIL
== bd053f0 (fpu_lib: decide FMA32 tininess by an f64 thre)
  F-fcvt.l.s-00 FAIL
  F-fcvt.s.w-00 FAIL
  F-fcvt.wu.s-00 FAIL
```

src/util/fpu_lib.h
SolAstrius: I tried exactly that and it mis-rounds — `sum` isn't exact (the product is, but `product + c` can span more bits than f64 holds), and downcasting the 53-bit rounding erases 24-bit tie-breaks:

```c
// cc -O2 tie_probe.c -lm && ./a.out
#include <stdio.h>
#include <stdint.h>
#include <string.h>
#include <math.h>
volatile float a, b, c;
int main(void)
{
    uint32_t ua = 0x6BE58000, ub = 0x3A123B00, uc = 0x117D0E3E, r1, r2;
    memcpy((void*)&a, &ua, 4); memcpy((void*)&b, &ub, 4); memcpy((void*)&c, &uc, 4);
    float f1 = fmaf(a, b, c), f2 = (float)((double)a * b + c);
    memcpy(&r1, &f1, 4); memcpy(&r2, &f2, 4);
    printf("fmaf     %08x\ndowncast %08x\n", r1, r2);
}
```
```
# arm64 and x86_64, identical:
fmaf     668317e5
downcast 668317e4
```

`a*b` sits exactly on a 24-bit tie and `c` breaks it upward; the intermediate rounding loses that. (2p+2 doesn't apply — the addends aren't 24-bit values.) So `fpu_fma32`'s own rounding keeps the odd-round; the PR text claiming I "fixed" it was wrong, rewording.

src/util/fpu_lib.h
SolAstrius: The premise fails: `fcvt.s.d` doesn't set UF correctly on aarch64 either. (You mean the narrowing `fpu_fcvt_f64_to_f32` — the widening one is exact.)

```c
// cc -O2 uf_probe.c -lm && ./a.out
// Every case rounds up from below onto exactly +min-normal:
// after-rounding tininess says NX only.
#include <stdio.h>
#include <stdint.h>
#include <string.h>
#include <math.h>
#include <fenv.h>
volatile float fa, fb, fr;
volatile double da, db, dr;
static void flags(const char* op, uint64_t bits)
{
    int e = fetestexcept(FE_ALL_EXCEPT);
    printf("%-8s %016llx  %s%s\n", op, (unsigned long long)bits,
           (e & FE_INEXACT) ? "NX " : "", (e & FE_UNDERFLOW) ? "UF" : "");
    feclearexcept(FE_ALL_EXCEPT);
}
int main(void)
{
    uint32_t a32 = 0x207FFFFE, b32 = 0x1F800001, r32; // 2^-62*(1-2^-23), 2^-64*(1+2^-23)
    uint64_t a64 = 0x200FFFFFFFFFFFFE, b64 = 0x1FF0000000000001, r64;
    memcpy((void*)&fa, &a32, 4); memcpy((void*)&fb, &b32, 4);
    memcpy((void*)&da, &a64, 8); memcpy((void*)&db, &b64, 8);
    feclearexcept(FE_ALL_EXCEPT);
    fr = fa * fb;         memcpy(&r32, (void*)&fr, 4); flags("mul.s", r32);
    fr = fmaf(fa, fb, 0); memcpy(&r32, (void*)&fr, 4); flags("fma.s", r32);
    dr = da * db;         memcpy(&r64, (void*)&dr, 8); flags("mul.d", r64);
    dr = fma(da, db, 0);  memcpy(&r64, (void*)&dr, 8); flags("fma.d", r64);
    dr = ldexp(1.0 - ldexp(1.0, -46), -126); feclearexcept(FE_ALL_EXCEPT);
    fr = (float)dr;       memcpy(&r32, (void*)&fr, 4); flags("cvt.s.d", r32);
}
```
```
# x86_64:                        # arm64 (M1):
mul.s    0000000000800000  NX    mul.s    0000000000800000  NX UF
fma.s    0000000000800000  NX    fma.s    0000000000800000  NX UF
mul.d    0010000000000000  NX    mul.d    0010000000000000  NX UF
fma.d    0010000000000000  NX    fma.d    0010000000000000  NX UF
cvt.s.d  0000000000800000  NX    cvt.s.d  0000000000800000  NX UF
```

ARM detects tininess before rounding on *every* instruction, x86 after. So nothing makes FMA special — mul/div/cvt all have the same bug; only add/sub are safe, since a sum near the boundary is always exact (shared quantum when both operands are small, Sterbenz's lemma on cancellation).

The right end state is fixing this once for all of them, agreed — but doing it correctly needs a flags snapshot from *before* the op, or a not-tiny verdict can't clear the bogus UF without also clearing a sticky one. fma already pays for that snapshot at entry; the hot ops don't. So: FMA here, cvt/mul/div/sqrt together as a follow-up.

SolAstrius:
Switched to the f64 compare in bd053f0; the per-mode thresholds are in the comment, the rest is in the two threads. One correction to the walk-through: `2^-126 - 2^-151` in the commit message is the *midpoint*, not the exact result (ambiguous wording, mine). The failing input sits just below it — e.g. `2^-126 - 2^-151 - 2^-197`, whose `sum` collapses onto the midpoint exactly — which is why comparing `sum` alone needs the TwoSum tie-break there, and only there. Done committing on this one.

src/cpu/riscv_fpu.c
```diff
@@ -81,8 +81,9 @@ static void riscv_prepare_rmm(rvvm_hart_t* vm, const uint32_t insn, const size_t
             neg = fpu_signbit64(riscv_view_d(vm, rs1)) != fpu_signbit64(riscv_view_d(vm, rs2));
             break;
         default:
-            neg = fpu_signbit64(riscv_view_d(vm, rs1));
-            break;
+            // Only add/sub/mul/div need the directed synthesis: sqrt has no exact
+            // ties, and ops taking rm as an argument handle RMM natively
```
purplesyringa, line 85: I'm probably misunderstanding something, but doesn't `sqrt(0.25)` yield an exact tie? Or are we just going to say this is broken?

src/util/fpu_lib.h
purplesyringa: Oh, now I see, `(float)(a * b + c)` has double rounding, one after `+` and one due to `(float)`, thanks. I'll re-review the code taking this into account in a bit.

src/cpu/riscv_fpu.c
SolAstrius: > I'm probably misunderstanding something, but doesn't `sqrt(0.25)` yield an exact tie? Or are we just going to say this is broken?

sqrt(0.25) = 0.5 is exact though?... nothing rounds, every mode returns it, RMM as well.

or are we using "tie" differently? I mean the exact result landing halfway between two representables — the only place RMM and RNE disagree. an exactly-representable result like 0.5 isn't a tie, there's nothing to round. and sqrt can't hit a midpoint: that would be an odd 25-bit significand whose square (odd, ~49 bits) isn't a valid 24-bit input. swept all positive f32 to be sure — 0 midpoints in 2139095039 inputs.

purplesyringa: Oh yeah, that makes more sense, thanks. The only edge case you didn't mention is subnormals, since those have more limited precision, but squares of (midpoints between) subnormals aren't representable in floats either, for separate reasons.

purplesyringa:
Okay. Now that I understand the issue better, I think I like the previous version of `fpu_fma32_fixup_uf` (the one with `fpu_odd_round64`, not with hard-coding subtle boundaries manually) more: its semantics look much more obvious if you know how FMA32 emulation works. So I'd appreciate it if you rolled back bd053f08e4efcb723884671a53ceabb9be403b49 -- sorry for asking you to do double work.

Building upon that: in the previous version, `fpu_fma32_fixup_uf` pretty much just repeats the work of `fpu_fma32`, except that it multiplies an intermediate value by `2^126` close to the end. That's all well and good if we don't have hardware FMA, because it allows `fpu_fma32_fixup_uf` to reuse results computed by `fpu_fma32` (if the optimizer is willing enough). But if we do have hardware FMA, UF handling becomes relatively expensive, since we have to recompute what is essentially just `fma32(a * 2^126, b, c * 2^126)` with a software method despite having hardware FMA at our disposal.

So I guess what I'm asking is: would it make sense to copy the approach used in `fpu_fma64_fixup_uf`, i.e. simply recomputing `fpu_fma32(...)` with values scaled by (I assume) `2^23`, if hardware FMA is available? The software-FMA path could instead reuse `res` from the software path in `fpu_fma32`. It's going to require some moving around (possibly inlining fixups into the FMA, or adding `fpu_fma32_raw`, or something), but I think it has a chance to be simpler than the current implementation.

purplesyringa:
I also have a slightly unrelated question, feel free to handle it elsewhere, but it's related to rounding modes that this PR fixes, so it would make just as much sense to resolve it here.

`fpu_fma32` and `fpu_fma64` can be run with any active rounding mode. The odd-rounding path works fine for round-to-nearest, but does it work for other roundings?

[The original paper](https://guillaume.melquiond.fr/doc/08-tc.pdf) covering our `fpu_fma32` links to [When double rounding is odd](https://inria.hal.science/inria-00070603v2/document), which says

> The two algorithms are even more general than what we presented here. They are not limited to a final result rounded to nearest even, they can also be applied to any other realistic rounding (meaning that the result of a computation is uniquely defined by the exact value of the real result and does not depend on the machine state). In particular, they handle all the IEEE-754 standard roundings and the new rounding to the nearest away from zero defined by the revision of this standard.

...so I'm assuming `fpu_odd_round64` is safe, but it says nothing of our `fpu_add_error64`, and [the paper the code links to](https://ir.cwi.nl/pub/9159/9159D.pdf) mentions that

> The theorems do not hold if addition [rounds arbitrarily], because the correction term zz is then not always an element of R.

They proceed to show an example where, in ceiling mode, `15 + 15/32` rounds to `16` with the error term `-17/32`. Here,

```
15     =  1.111 * 2^3
15/32  =  1.111 * 2^-2
-17/32 = -1.0001 * 2^-1
```

...so the error term requires more mantissa bits than both inputs, and thus it's not necessarily representable in the same floating-point type. I think this counterexample can be extended to our floats/doubles as well, which means we might compute `err` incorrectly. I don't know if this can change the sign of `err`, which is the only thing we rely on, but this feels messy, and I'm not sure how to verify it.

So: would you be able to run the FPU test suite with hardware FMA disabled and a non-NE rounding mode and verify if any tests fail, or try to manufacture/brute-force a failure? If that uncovers a problem, as I suspect it will, we'll need to change the way we handle FMA roundings.

I *think* that double rounding is innocuous for flooring and ceiling, so e.g. `(float)((double)a * (double)b + (double)c)` would compute 32-bit FMA exactly right in those modes. Truncation should also be fine because it's equivalent either to flooring or ceiling. That fix should be as simple as adding an `if` around the error correction part.

I'm not so sure what to do about the 64-bit case, though, since we'll still be rounding after multiplication, and that can break catastrophic cancellation...

LekKit:
Follow-up discussion at https://github.com/LekKit/RVVM/discussions/250

PR #259, opened 2026-08-11
fpu: canonicalize NaN results of fadd/fmul/fdiv and fused ops

### LINKED ISSUE #254: `fadd`/`fmul`/`fdiv` leak qNaN payloads instead of returning canonical NaNs
## Summary

`fadd`, `fmul` and `fdiv` preserve the input qNaN payload in the result instead of returning the canonical NaN required by the RISC-V specification. `fsub` returns a canonical NaN in the same probe, which shows this is a shared writeback-path problem rather than a global NaN-handling gap.

## Expected behavior

Native RISC-V hardware and QEMU return the canonical NaN for `fadd`, `fsub`, `fmul`, `fdiv` and `fsqrt`:

```text
single: 0xffffffff7fc00000
double: 0x7ff8000000000000
```

## Observed behavior

```text
single: fadd=0xffffffff7fc01234 fsub=0xffffffff7fc00000
        fmul=0xffffffff7fc01234 fdiv=0xffffffff7fc01234
double: fadd=0x7ff8000000001234 fsub=0x7ff8000000000000
        fmul=0x7ff8000000001234 fdiv=0x7ff8000000001234
```

## Root cause

In `src/cpu/riscv_fpu.c`, `fadd`/`fmul`/`fdiv` use the `riscv_emit_s`/`riscv_emit_d` write path, while `fsub` uses `riscv_write_s`/`riscv_write_d`. The write helpers canonicalize NaNs before writeback; the emit helpers do not.

## Suggested fix

Route the arithmetic writebacks through the same canonicalizing writers used by `fsub`.

No existing report for this behavior was found in the issue tracker.

carlosqwqqwq:
# fpu: canonicalize NaN results of fadd/fmul/fdiv and fused ops

Fixes #254

## Description

The FP writeback for `fadd`, `fmul`, `fdiv` and the fused `fmadd/fmsub/fnmsub/fnmadd` family stores the raw host result, which keeps non-canonical NaN payloads. The `fsub`/`fmin`/`fmax` paths already canonicalize through the write helpers; the RISC-V specification requires the canonical NaN bit pattern for arithmetic results. Route the remaining arithmetic writebacks through the same canonicalizing writers.

Signaling-NaN `NV` flagging is unchanged.

## Validation

- qNaN-payload, `inf*0+1` and `0/0` probes match native RISC-V hardware and QEMU after the fix, with the JIT enabled and disabled.
- Every arithmetic result is the canonical NaN and `fflags` match the references.

### DIFF
--- a/src/cpu/riscv_fpu.c
+++ b/src/cpu/riscv_fpu.c
@@ -140,10 +140,10 @@ static slow_path void riscv_emulate_f_opc_op_impl(rvvm_hart_t* vm, const uint32_
              * FPU computations
              */
             case RISCV_FPU_GEN_RM_CASES(0x00000000UL): // fadd.s
-                riscv_emit_s(vm, rds, fpu_add32(riscv_read_s(vm, rs1), riscv_read_s(vm, rs2)));
+                riscv_write_s(vm, rds, fpu_add32(riscv_read_s(vm, rs1), riscv_read_s(vm, rs2)));
                 return;
             case RISCV_FPU_GEN_RM_CASES(0x02000000UL): // fadd.d
-                riscv_emit_d(vm, rds, fpu_add64(riscv_view_d(vm, rs1), riscv_view_d(vm, rs2)));
+                riscv_write_d(vm, rds, fpu_add64(riscv_view_d(vm, rs1), riscv_view_d(vm, rs2)));
                 return;
             case RISCV_FPU_GEN_RM_CASES(0x08000000UL): // fsub.s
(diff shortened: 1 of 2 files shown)


LekKit:
@SolAstrius please confirm via your out-of-tree test suite and review if possible, thank you

carlosqwqqwq:
The affected fadd/fmul/fdiv writebacks now match the canonical NaN reference. fsqrt remains outside this PR and is tracked separately. The PR checks pass; please confirm with your independent suite when convenient.

PR #280, opened 2026-08-12
riscv: preserve fflags when rejecting a reserved frm value

### LINKED ISSUE #274: Writing a reserved frm value through fcsr silently drops the fflags update
# Writing a reserved `frm` value through `fcsr` silently drops the `fflags` update

## Summary

In the current RVVM revision, a guest write to `fcsr` that carries a reserved rounding-mode value in the `frm` field (101/110/111) is rejected as a whole: `riscv_update_fcsr()` returns early and the `fflags` bits written by the same instruction are lost. On real RV64 hardware and in QEMU, the `fflags` field update takes effect and only the reserved `frm` value is handled per WARL semantics (kept or mapped to a supported value). The bug silently corrupts floating-point exception state restoration for any software that writes `fcsr` with a non-canonical `frm` value.
## Affected revision

2026-09-06.
## Reproduction

Build a minimal RV64 guest probe with the RISC-V GNU toolchain
(`riscv64-linux-gnu-gcc -static -O2 -march=rv64gc`). The probe performs, among
others, this sequence:

```c
csr_write(FCSR, 0);
csr_write(FCSR, (5 << 5) | 0x1f);   /* reserved frm=5, all fflags set */
printf("fcsr=%08" PRIx64 "...", csr_read(FCSR));

```

## Observed behavior

Real RV64 hardware (native RISC-V hardware, Linux 6.6.63) and QEMU print:

```text
case2_frm5 fcsr=000000bf frm=5 fflags=1f

```

The current RVVM build prints:

```text
case2_frm5 fcsr=00000000 frm=0 fflags=0

```

RVVM drops the complete write, including the `fflags` update. The same behavior occurs for `frm=6` and `frm=7`, and for a direct `frm` CSR write of a reserved value (where QEMU/hardware keep the written value; WARL permits either keeping or mapping it).
## Root cause

`src/cpu/riscv_csr.c`, `riscv_update_fcsr()`:

```c
if (unlikely(new_frm != old_frm)) {
if (new_frm > RM_RMM) {
    // Invalid rounding mode written: keep the old FRM value
    new_frm = old_frm;
    } else {
        // Set host rounding mode
        fpu_set_rounding_mode(new_frm);
    }
}

```

The early `return` aborts the whole CSR write. `fcsr` is a single CSR whose `fflags` field (bits 4:0) is a plain read/write field; only the `frm` field (bits 7:5) is WARL. The `fflags` update must be applied regardless of the reserved `frm` value.
## Minimal fix

Keep the old `frm` value (WARL) and continue applying the remaining fields:

```c
if (new_frm > RM_RMM) {
    // Invalid rounding mode written: keep the old FRM value per
    // WARL semantics, but still apply the other fields of fcsr
    new_frm  = old_frm;
} else {
    // Set host rounding mode
    fpu_set_rounding_mode(new_frm);
}

vm->csr.fcsr = new_fflags | (new_frm << 5);

```

The proposed change contains this change. In an isolated source copy of the affected revision, the patched build prints:

```text
case2_frm5 fcsr=0000001f frm=0 fflags=1f

```

(`fflags` applied; reserved `frm` mapped to the previous value, which is the WARL-preserving choice).
## Specification references

- RISC-V unpriv spec, F extension: `fcsr`/`frm`/`fflags` field layout and reserved rounding-mode encodings 101-110 (static) and 101-111 (dynamic).
- RISC-V priv spec, CSR field semantics: WARL fields accept any written value and return a legal value on read; only the WARL field itself is affected by an unsupported write.

carlosqwqqwq:
# riscv: preserve fflags when rejecting a reserved frm value

Fixes #274

## Commit message

```text
riscv: preserve fflags when rejecting a reserved frm value

```

## Description

`riscv_update_fcsr()` currently returns before committing the writable `fflags` field when the incoming `frm` value is reserved. Keep the old/WARL `frm` value, apply the independent `fflags` bits, and reconstruct the final CSR from the validated fields. The change is limited to `src/cpu/riscv_csr.c` and applies independently to `staging`; validation artifacts stay out of the source diff.
## Validation

### DIFF
--- a/src/cpu/riscv_csr.c
+++ b/src/cpu/riscv_csr.c
@@ -425,8 +425,8 @@ static void riscv_update_fcsr(rvvm_hart_t* vm, uint32_t old_fcsr, uint32_t new_f
         uint32_t new_fflags = bit_cut(new_fcsr, 0, 5);
         if (unlikely(new_frm != old_frm)) {
             if (new_frm > RM_RMM) {
-                // Invalid rounding mode written
-                return;
+                // Keep invalid frm, but still apply the other fcsr fields
+                new_frm = old_frm;
             } else {
                 // Set host rounding mode
                 fpu_set_rounding_mode(new_frm);
@@ -438,7 +438,7 @@ static void riscv_update_fcsr(rvvm_hart_t* vm, uint32_t old_fcsr, uint32_t new_f
                 fpu_set_exceptions(0);
             }
         }
-        vm->csr.fcsr = new_fcsr;
+        vm->csr.fcsr = new_fflags | (new_frm << 5);
     }
 }
 


src/cpu/riscv_csr.c
```diff
@@ -425,8 +425,10 @@ static void riscv_update_fcsr(rvvm_hart_t* vm, uint32_t old_fcsr, uint32_t new_f
         uint32_t new_fflags = bit_cut(new_fcsr, 0, 5);
         if (unlikely(new_frm != old_frm)) {
             if (new_frm > RM_RMM) {
-                // Invalid rounding mode written
-                return;
+                // Invalid rounding mode written: keep the old FRM value per
+                // WARL semantics, but still apply the other fields of fcsr
+                new_fcsr = bit_replace(new_fcsr, 5, 3, old_frm);
```
purplesyringa: Maybe reconstruct `new_fcsr` from the patched `new_frm`/`new_fflags` once at writeback so that this isn't a path uncovered by most tests?

purplesyringa (changes requested):

src/cpu/riscv_csr.c
carlosqwqqwq: Thanks! Reworked: fcsr is now reconstructed from the patched frm/fflags once at writeback (vm->csr.fcsr = (new_fcsr & ~CSR_FCSR_MASK) | new_fflags | (new_frm << 5)), so the reserved-frm path shares the same writeback as every other path. Rebuilt cleanly and force-pushed.

carlosqwqqwq:
Reworked the reserved-frm path: an invalid frm keeps the old value, while the final fcsr writeback applies fflags and the validated frm together. CI passes; please re-review.

src/cpu/riscv_csr.c
```diff
@@ -438,7 +439,9 @@ static void riscv_update_fcsr(rvvm_hart_t* vm, uint32_t old_fcsr, uint32_t new_f
                 fpu_set_exceptions(0);
             }
         }
-        vm->csr.fcsr = new_fcsr;
+        // Reconstruct fcsr from the patched frm/fflags so every path shares
+        // the same writeback (reserved frm keeps the old value via WARL)
+        vm->csr.fcsr = (new_fcsr & ~CSR_FCSR_MASK) | new_fflags | (new_frm << 5);
```
purplesyringa, line 442: Is `(new_fcsr & ~CSR_FCSR_MASK) |` necessary here? I would assume all of those bits are either zero or zeroed on read, would you be able to verify whether that's the case?

carlosqwqqwq: Yes, this mask is necessary for architectural compatibility.

The current RISC-V F extension specification defines `fcsr[31:8]` as reserved for other standard extensions and explicitly says that standard software should preserve those bits. An implementation without those extensions may read them as zero, but RVVM must not discard them during a read-modify-write path when a future/implemented extension uses them. `CSR_FCSR_MASK` is `0x000000ff`, so:

`(new_fcsr & ~CSR_FCSR_MASK) | new_fflags | (new_frm << 5)`

preserves the non-F-extension portion while replacing only `frm` and `fflags` with their validated values. I also removed the misleading WARL wording from the comments; the code preserves RVVM's existing invalid-`frm` policy while still applying the remaining fields.

The updated head is `18014ab`. It builds successfully with `make -j2` in the x86_64 lab container.

Reference: https://github.com/riscv/riscv-isa-manual/blob/main/src/unpriv/f-st-ext.adoc

carlosqwqqwq:
Applied the follow-up from your inline review: the final writeback now uses the validated fflags and frm fields directly, with no duplicate CSR_FCSR_MASK preservation at this point. The branch is now a single commit based on current staging; local build and patch checks pass. Please re-review.

src/cpu/riscv_csr.c
purplesyringa: We already mask it [here](https://github.com/LekKit/RVVM/blob/e8fa67586b14e4eb01992956b653c326f87d3853/src/cpu/riscv_csr.c#L480) though, so only handling the mask in this location doesn't look sufficient. Regardless, I think the intention of `CSR_FCSR_MASK` is to contain all important bits, including those provided by extensions, so keeping bits outside of the mask is useless anyway.

It's reasonable to ask if this snippet can be written in such a way that it won't have to be updated if extensions are implemented, but since this is *the* function that handles FCSR changes, that looks like exactly the place that should worry about the entirety of defined bits. So I think it's fine to just drop the masking here, under the assumption that if a new extension is added, this function is going to be the first place that we'll need to touch anyway.

src/cpu/riscv_csr.c
```diff
@@ -425,8 +425,8 @@ static void riscv_update_fcsr(rvvm_hart_t* vm, uint32_t old_fcsr, uint32_t new_f
         uint32_t new_fflags = bit_cut(new_fcsr, 0, 5);
         if (unlikely(new_frm != old_frm)) {
             if (new_frm > RM_RMM) {
-                // Invalid rounding mode written
-                return;
+                // Keep invalid frm, but still apply the other fcsr fields
```
LekKit:
```
// Ignore invalid frm, apply other fcsr fields
```

PR #281, opened 2026-08-12
riscv: sign-extend operands in signed byte atomics

### LINKED ISSUE #276: RV64 amomin.b/amomax.b (Zabha) treat the memory byte as unsigned
# [RV64] `amomin.b`/`amomax.b` (Zabha) treat the memory byte as unsigned

## Summary

RVVM's implementation of the Zabha byte min/max atomics compares the memory byte as an *unsigned* 0..255 value instead of a signed -128..127 value, and the `rs2` operand is passed the same way. For any byte with bit 7 set, `amomin.b` and `amomax.b` therefore store (or keep) the opposite operand of what the RISC-V specification requires. Concretely, with the memory byte `0x80` and `rs2 = 1`:
- `amomax.b` must store `1` (max(-128, 1) = 1); RVVM keeps `0x80`.
- `amomin.b` must keep `0x80` (min(-128, 1) = -128); RVVM stores `1`. And with the memory byte `0x00` and `rs2 = 0xFF` (-1):
- `amomax.b` must keep `0x00` (max(0, -1) = 0); RVVM stores `0xFF`.
- `amomin.b` must store `0xFF` (min(0, -1) = -1); RVVM keeps `0x00`. The signed 16-bit variants (`amomin.h`/`amomax.h`) and the unsigned byte variants (`amominu.b`/`amomaxu.b`) behave correctly.
## Reproducer

The minimal probe source executes each case as a raw-encoded instruction with known register and memory state and prints the resulting `rd` and memory word. Build it for RV64 with the RISC-V GNU toolchain: `riscv64-linux-gnu-gcc -static -O2 -march=rv64gc`. Reference output (QEMU with Zacas/Zabha/Zicboz enabled, `qemu-riscv64 -cpu max`):

```text
zabha_amomin_b    rd=0xffffffffffffff80 mem=0x0000000000000080
zabha_amomax_b    rd=0xffffffffffffff80 mem=0x0000000000000001
zabha_amomin_b_neg rd=0x0000000000000000 mem=0x00000000000000ff
zabha_amomax_b_neg rd=0x0000000000000000 mem=0x0000000000000000

```

RVVM output (identical with JIT enabled and disabled):

```text
zabha_amomin_b    rd=0xffffffffffffff80 mem=0x0000000000000001
zabha_amomax_b    rd=0xffffffffffffff80 mem=0x0000000000000080
zabha_amomin_b_neg rd=0x0000000000000000 mem=0x0000000000000000
zabha_amomax_b_neg rd=0x0000000000000000 mem=0x00000000000000ff

```

The old-value register write (`rd`) is correct (sign-extended); only the comparison and the store are wrong.
## Root cause

Two places treat the byte operands as unsigned:
- `src/util/atomics.h`, `atomic_max_int8()`/`atomic_min_int8()` load the memory byte with `atomic_load_uint8()` into an `int16_t`; values 0x80..0xFF stay 128..255 instead of becoming -128..-1, so the signed comparison `tmp > val` / `tmp < val` selects the wrong side.
- `src/cpu/riscv_atomics.h`, `riscv_emulate_atomic_b()` reads the `rs2` low byte as `uint8_t` and passes it to those helpers; 0x80..0xFF are again compared as 128..255. The 16-bit paths are unaffected only because the `uint16_t` to `int16_t` conversion on this host wraps to the signed value.
## Suggested fix

Sign-extend both operands in the signed-byte helper:

```c
/* src/util/atomics.h */
tmp = (int16_t)(int8_t)atomic_load_uint8(addr);

/* src/cpu/riscv_atomics.h */
const int16_t val = (int16_t)(int8_t)riscv_read_reg(vm, rs2);

```

The patched build matches the reference on the complete 38-case matrix (including all signed/unsigned 8/16-bit min/max boundary cases) with both JIT and interpreter paths.

carlosqwqqwq:
# riscv: sign-extend operands in signed byte atomics

Fixes #276

## Commit message

```text
riscv: sign-extend operands in signed byte atomics

```

## Description

The signed `amomin.b`/`amomax.b` emulation compares zero-extended bytes. Sign-extend both the memory byte and the low byte of `rs2` before entering the signed comparison helper. The change is limited to `src/util/atomics.h` and `src/cpu/riscv_atomics.h`; validation artifacts stay out of the source diff.
## Validation

### DIFF
--- a/src/cpu/riscv_atomics.h
+++ b/src/cpu/riscv_atomics.h
@@ -42,7 +42,7 @@ static forceinline void riscv_emulate_atomic_b(rvvm_hart_t* vm, const uint32_t i
     const size_t   rs1   = bit_ext_u32(insn, 15, 5);
     const size_t   rs2   = bit_ext_u32(insn, 20, 5);
     const xaddr_t  vaddr = riscv_read_reg(vm, rs1);
-    const uint8_t  val   = riscv_read_reg(vm, rs2);
+    const int16_t  val   = (int16_t)(int8_t)riscv_read_reg(vm, rs2);
     uint8_t        buf   = 0;
     void*          ptr   = NULL;
 
--- a/src/util/atomics.h
+++ b/src/util/atomics.h
@@ -1967,7 +1967,7 @@ static inline int16_t atomic_max_int8(void* addr, int16_t val)
 {
     int16_t tmp;
     do {
-        tmp = atomic_load_uint8(addr);
+        tmp = (int16_t)(int8_t)atomic_load_uint8(addr);
     } while (!atomic_cas_uint8(addr, tmp, tmp > val ? tmp : val));
     return tmp;
 }
@@ -1976,7 +1976,7 @@ static inline int16_t atomic_min_int8(void* addr, int16_t val)
 {
     int16_t tmp;
     do {
-        tmp = atomic_load_uint8(addr);
+        tmp = (int16_t)(int8_t)atomic_load_uint8(addr);
     } while (!atomic_cas_uint8(addr, tmp, tmp < val ? tmp : val));
     return tmp;
 }


LekKit:
I understand the goal, but I think 84417f0 is a more suitable fix for this (Among few other issues in atomics helper).
Please verify that, if it works you can close this PR and related issue.

carlosqwqqwq:
Verified: 84417f0 (staging head) already fixes the root cause. After the helper types were changed to int8_t/uint8_t, both the memory byte and the rs2 operand are converted to signed int8_t at the call sites (uint8_t -> int8_t implicit conversion), so amomin.b/amomax.b now compare as signed, matching Zabha. Our patch is therefore redundant. Closing this PR and the related issue as you suggested. Thanks for the fix!

PR #282, opened 2026-08-12
riscv: include negative minimum values in FCVT fit checks

### LINKED ISSUE #277: fcvt.w/l.* set the invalid flag for exact negative boundary inputs
# `fcvt.w/l.*` set the invalid flag for exact negative boundary inputs

## Affected revision

2026-09-06; x86-64 host, RV64 guest user-mode (JIT and interpreter both affected).
## Summary

The "fits" helpers used by the float-to-integer conversions compare the magnitude of the input against the *exclusive* limit of the destination type:

```c
static forceinline bool fpu_f32_fits_i32(fpu_f32_t f)
{
    return (fpu_bit_f32_to_u32(f) << 1) < 0x9E000000U;   // |f| < 2^31
}

```

`-2^31` is exactly representable and is the minimum value of `int32`, so `fcvt.w.s(-2^31)` must convert with no exception flags. The strict `<` classifies it as out-of-range, and the conversion path raises the invalid flag (`fflags.NV`, `0x10`). The returned value happens to be correct (`0x80000000`, which is also the negative saturation value), so only the flag is wrong. The same applies to `fcvt.w.d(-2^31)` and, for the 64-bit conversions, to `fcvt.l.s(-2^63)` / `fcvt.l.d(-2^63)`.
## Test cases

| Case | native RISC-V hardware / QEMU | RVVM (JIT & interpreter) |
| --- | --- | --- |
| `fcvt.w.s(-2^31)` | `0xffffffff80000000`, flags=`0` | `0xffffffff80000000`, flags=`0x10` (wrong) |
| `fcvt.w.d(-2^31)` | `0xffffffff80000000`, flags=`0` | `0xffffffff80000000`, flags=`0x10` (wrong) |
| `fcvt.l.s(-2^63)` | `0x8000000000000000`, flags=`0` | `0x8000000000000000`, flags=`0x10` (wrong) |
| `fcvt.l.d(-2^63)` | `0x8000000000000000`, flags=`0` | `0x8000000000000000`, flags=`0x10` (wrong) |
Control cases behave identically on all engines: `fcvt.w.s(-2^31-1024)` (overflow) and `fcvt.w.s(2^31)` (overflow) give `flags=0x10` everywhere; the boundary inputs are the only disagreement.
## Root cause

`src/util/fpu_lib.h`, the four `fpu_*_fits_i32` / `fpu_*_fits_i64` helpers:
- `fpu_f32_fits_i32`: `(u << 1) < 0x9E000000U` (must also accept `0xCF000000` = `-2^31`)
- `fpu_f64_fits_i32`: `(u << 1) < 0x83C0000000000000ULL` (must also accept `0xC1E0000000000000`)
- `fpu_f32_fits_i64`: `(u << 1) < 0xBE000000U` (must also accept `0xDF000000` = `-2^63`)
- `fpu_f64_fits_i64`: `(u << 1) < 0x87C0000000000000ULL` (must also accept `0xC3E0000000000000`)
## Minimal fix

The proposed patch adds the exact negative-boundary bit patterns to each fits predicate (positive boundaries remain exclusive, as `+2^31` and `+2^63` overflow the destination):

```c
static forceinline bool fpu_f32_fits_i32(fpu_f32_t f)
{
    uint32_t u = fpu_bit_f32_to_u32(f);
    return (u << 1) < 0x9E000000U || u == 0xCF000000U;
}

```

carlosqwqqwq:
# riscv: include negative minimum values in FCVT fit checks

Fixes #277

## Commit message

```text
riscv: include negative minimum values in FCVT fit checks

```

## Description

The float-to-integer fit predicates use a strict magnitude comparison and classify exactly representable `-2^31`/`-2^63` inputs as invalid. Include the negative minimum bit patterns while retaining the positive overflow checks. The change is limited to the four fit predicates in `src/util/fpu_lib.h`; validation artifacts stay out of the source diff.
## Validation

### DIFF
--- a/src/util/fpu_lib.h
+++ b/src/util/fpu_lib.h
@@ -12,6 +12,7 @@ file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 #include "fpu_types.h"
 #include "mem_ops.h"
+#include "bit_ops.h"
 
 /*
  * NOTE: Floating-point is actually subtly (or seriously) broken on many compilers and platforms
@@ -1357,12 +1358,14 @@ static forceinline bool fpu_f64_fits_u32(fpu_f64_t d)
 
 static forceinline bool fpu_f32_fits_i32(fpu_f32_t f)
 {
-    return (fpu_bit_f32_to_u32(f) << 1) < 0x9E000000U;
+    uint32_t u = fpu_bit_f32_to_u32(f);
+    return (bit_rotl32(u, 1) ^ 1) <= 0x9E000000U; // |f| < 2^31 or f == -2^31
 }
 
 static forceinline bool fpu_f64_fits_i32(fpu_f64_t d)
 {
-    return (fpu_bit_f64_to_u64(d) << 1) < 0x83C0000000000000ULL;
+    uint64_t u = fpu_bit_f64_to_u64(d);
+    return (bit_rotl64(u, 1) ^ 1) <= 0x83C0000000000000ULL; // |d| < 2^31 or d == -2^31
 }
 
 static forceinline bool fpu_f32_fits_u64(fpu_f32_t f)
@@ -1379,12 +1382,14 @@ static forceinline bool fpu_f64_fits_u64(fpu_f64_t d)
 
 static forceinline bool fpu_f32_fits_i64(fpu_f32_t f)
 {
-    return (fpu_bit_f32_to_u32(f) << 1) < 0xBE000000U;
+    uint32_t u = fpu_bit_f32_to_u32(f);
+    return (bit_rotl32(u, 1) ^ 1) <= 0xBE000000U; // |f| < 2^63 or f == -2^63
 }
 
 static forceinline bool fpu_f64_fits_i64(fpu_f64_t d)
 {
-    return (fpu_bit_f64_to_u64(d) << 1) < 0x87C0000000000000ULL;
+    uint64_t u = fpu_bit_f64_to_u64(d);
+    return (bit_rotl64(u, 1) ^ 1) <= 0x87C0000000000000ULL; // |d| < 2^63 or d == -2^63
 }
 
 /*


purplesyringa:
You can speed up this check with something like

```c
return (bit_rotl32(f, 1) ^ 1) <= 0x9E000000U;
```

Technically you can remove yet another instruction on x86 by going

```asm
shl f, 1
sbc f, 0x9E000000
jb ...
```

but it's messy in C, can't be implemented efficiently on ARM, and is deoptimized by GCC, so the `rotl` approach is probably the best.

```c
_Bool fpu_f32_fits_i32_yuki(unsigned f) {
    unsigned twof;
    _Bool carry_in = __builtin_add_overflow(f, f, &twof);
    unsigned carry_out;
    __builtin_subc(twof, 0x9E000000U, carry_in, &carry_out);
    return carry_out;
}
```

carlosqwqqwq:
Thanks for the suggestion! Adopted the rotate-based check in all four fits helpers: (bit_rotl32(u, 1) ^ 1) <= 0x9E000000U for f32/i32, and the rotl64 counterparts for f64/i32 and the i64 pairs. Verified bit-exact equivalence against the previous two-comparison form over the full exponent/mantissa grids (zero mismatches), rebuilt cleanly, and force-pushed.

carlosqwqqwq:
I adopted the rotate-based check at head `73e78e1a` in all four signed fit helpers (`f32/f64` to `i32/i64`). I verified bit-exact equivalence against the previous two-comparison form over the full exponent/mantissa test grids with zero mismatches, and all current GitHub Actions checks pass. @purplesyringa, please re-review.

purplesyringa (approved):

carlosqwqqwq:
I rechecked the current PR head `73e78e1` against the maintainer's rotate-based suggestion. The source already contains the exact `bit_rotl32/64(u, 1) ^ 1` implementation in all four signed fit helpers, including the explicit `bit_ops.h` include, so there is no additional code change to push and no empty commit was created.

Validation completed on this PR head:
- source diff is clean;
- the RVVM x86_64 lab-container build succeeds with `make -j2`;
- the resulting `release.linux.x86_64/rvvm_x86_64` binary is present and executable.

The implementation accepts the exact negative limits (`-2^31` and `-2^63`) while retaining the existing rejection of out-of-range values. Please re-review the already-updated head when convenient.

PR #286, opened 2026-08-12
riscv: trap reserved C.LUI/C.ADDI16SP encodings instead of mis-decoding them

### LINKED ISSUE #284: RVVM executes reserved C.LUI/C.ADDI16SP and C.ADDIW rd=x0 encodings
# RVVM executes reserved C.LUI/C.ADDI16SP and C.ADDIW rd=x0 encodings

## Summary

The current RVVM master revision (as of 2026-08-13) decodes reserved 16-bit encodings in
the `c.lui` / `c.addi16sp` opcode slot as `c.srli`/`c.srai`/`c.andi`/`c.sub`-family
instructions and executes them, instead of raising an illegal-instruction trap. The bug
affects both RV64 and RV32 guests, in the interpreter as well as the JIT backend.

## Affected encodings

The reserved encodings are `c.lui`/`c.addi16sp` with `nzimm == 0` that are not valid
`c.mop.N` (Zcmop) code points:

```
opcode: C1 quadrant (bits [1:0] == 01), funct3 == 3 (bits [15:13] == 011)
imm == 0: bit12 == 0 and bits [6:2] == 0
rd (bits [11:7]) is even (0..30) or odd above 15 (17..31)
```

24 encodings total: `0x6001 0x6101 0x6201 0x6301 0x6401 0x6501 0x6601 0x6701
0x6801 0x6881 0x6901 0x6981 0x6a01 0x6a81 0x6b01 0x6b81 0x6c01 0x6c81 0x6d01
0x6d81 0x6e01 0x6e81 0x6f01 0x6f81`.

Per the RISC-V "C" extension specification, `c.lui`/`c.addi16sp` require a non-zero
immediate (`nzimm`); a zero immediate (except for `c.mop.N` with `N in {1,3,...,15}`,
Zcmop) is a reserved encoding and must raise an illegal-instruction exception.

The same fall-through also affects the RV64-only `c.addiw` with `rd == x0`. Per the
specification these code points are also reserved ("C.ADDIW is valid only when
rd != x0"), so they must raise an illegal-instruction exception too. RVVM's
fall-through decodes them as `c.srli x0, x0, shamt`, which is also a no-op, so the
observable behavior is accidentally benign, but the decode path is wrong and emits a
spurious `c.srli` JIT trace.

## Root cause

In `riscv_emulate_c_c1()` (`src/cpu/riscv_compressed.h`), the `case 0x03` handler for
`c.lui`/`c.addi16sp`/`c.mop` ends with a plain `break;` when the encoding is reserved.
The `break` exits the `funct3` switch and execution falls through into the `c.misc-alu`
decoder (`c.srli`/`c.srai`/`c.andi`/`c.sub`/...), which then executes the reserved
encoding as if it were a valid `c.srli`-family instruction. The same defect exists in
the RV64 branch of `case 0x01` (`c.addiw` with `rd == x0`).

## Reproduction

Minimal probe (prints "R" iff the reserved encoding executes):

```asm
.option norvc
.section .text
.globl _start
_start:
    .half 0x6001   # c.lui x0, 0 (reserved: nzimm == 0)
    .half 0x0001
    nop
    li a0, 1
    la a1, msg
    li a2, 1
    li a7, 64
    ecall
exit:
    li a7, 93
    ecall
msg: .ascii "R"
```

Observed on a real RV64 RISC-V host (Linux user mode) with the Linux-userland shim:

| runner | result |
| --- | --- |
| native RV64 hardware | SIGILL (exit 132) |
| QEMU `-cpu max` (riscv64) | SIGILL (exit 132) |
| RVVM jit (before fix) | executes, prints "R" |
| RVVM nojit (before fix) | executes, prints "R" |
| RVVM jit/nojit (after fix) | illegal-instruction trap |

All 24 reserved encodings behave identically; after the fix all 24 trap on both the
JIT and interpreter paths, matching the hardware and QEMU. The `c.addiw x0, imm`
reserved encodings also trap after the fix, matching the spec and QEMU
(`c64_illegal`).

## Impact

Reserved-instruction semantics are part of the ISA contract: a valid implementation
must trap. Software (including security hardening and fault-injection harnesses) that
relies on reserved-encoding traps will silently behave differently under RVVM. The
bug is also reachable from untrusted guest code, e.g. through `c.lui x0, 0` used as a
`c.mop` probe on implementations without Zcmop.

## Suggested fix

Raise an illegal-instruction trap instead of falling through into the `c.misc-alu`
decoder (see `fix.patch`):

```c
            } else if ((rds & 0x01) && !(rds & 0x10)) { // c.mop.{1,15}
                return;
            }
            // Reserved: c.lui/c.addi16sp with nzimm == 0 (and not a valid c.mop)
            riscv_illegal_insn(vm, insn);
            return;
```

and the equivalent change for the RV64 `c.addiw` `rd == x0` case (raise an
illegal-instruction trap instead of falling through).

carlosqwqqwq:
# Fix: trap reserved C.LUI/C.ADDI16SP and C.ADDIW rd=x0 encodings

## Problem

`riscv_emulate_c_c1()` in `src/cpu/riscv_compressed.h` falls through from the
reserved `c.lui`/`c.addi16sp`/`c.mop` code points into the `c.misc-alu`
decoder, executing reserved encodings as `c.srli`/`c.srai`/`c.andi`/`c.sub`-
family instructions. The RV64 `c.addiw` `rd == x0` code points are also
reserved per the spec ("valid only when rd != x0") and fell through the same
way.

## Change

Both affected paths now return explicitly instead of falling through:

- `case 0x01` (RV64): `c.addiw` with `rd == x0` (reserved) -> `riscv_illegal_insn()`
- `case 0x03`: `c.lui`/`c.addi16sp` with `nzimm == 0` that is not a valid `c.mop.N`
  -> `riscv_illegal_insn()`

## Verification

- 24 reserved `c.lui`-slot encodings: trap on native RISC-V hardware (SIGILL), QEMU
  (SIGILL), and now RVVM (both JIT and interpreter).
- RV64 `c.addiw x0, imm` reserved encodings: now raise an illegal-instruction
  trap, matching the spec and QEMU (`c64_illegal`).
- Regression: `c.lui x0, 1` hint, `c.mop.N` (Zcmop), `c.addiw` with `rd != 0`,
  and all other C1 encodings still execute correctly.

Fixes #284.

### DIFF
--- a/src/cpu/riscv_compressed.h
+++ b/src/cpu/riscv_compressed.h
@@ -252,7 +252,10 @@ static forceinline void riscv_emulate_c_c1(rvvm_hart_t* vm, const uint32_t insn)
                 riscv_write_reg(vm, rds, (int32_t)(riscv_read_reg(vm, rds) + imm));
                 return;
             }
-            break;
+            // c.addiw with rd == x0 is a reserved encoding (spec: valid only
+            // when rd != x0); trap instead of falling through or executing
+            riscv_illegal_insn(vm, insn);
+            return;
 #else
             const xlen_t  pc  = riscv_read_reg(vm, RISCV_REG_PC);
             const int32_t off = decode_c_jal_imm(insn);
@@ -286,7 +289,9 @@ static forceinline void riscv_emulate_c_c1(rvvm_hart_t* vm, const uint32_t insn)
             } else if ((rds & 0x01) && !(rds & 0x10)) { // c.mop.{1,15}
                 return;
             }
-            break;
+            // Reserved: c.lui/c.addi16sp with nzimm == 0 (and not a valid c.mop)
+            riscv_illegal_insn(vm, insn);
+            return;
         }
         case 0x05: { // c.j
             const int32_t off = decode_c_jal_imm(insn);


LekKit:
How come is `c.addiw` with `rd = 0` a valid instruction? The documentation says:
```
C.ADDIW is only valid when rd≠x0; the code points with rd=x0 are reserved.
```
https://docs.riscv.org/reference/isa/v20260120/unpriv/c-st-ext.html#27-1-5-2-integer-register-immediate-operations

LekKit:
Other than that, the fix is sound. Previously both cases fell through into `riscv_illegal_insn(vm, insn);`, but the interpreter code was moved around and now incorrectly decodes them

carlosqwqqwq:
You are right - thank you for the catch. I checked the spec: C.ADDIW is valid only when rd != x0; the rd == x0 code points are reserved (not HINTs), and QEMU decodes them as c64_illegal. I have updated the patch so c.addiw rd == x0 now raises an illegal-instruction trap instead of being treated as a no-op, and force-pushed. The c.lui/c.addi16sp nzimm == 0 slot (the main subject of the PR) is unchanged.

PR #299, opened 2026-09-05
csr: validate CSRRS/CSRRC writes by register index

### LINKED ISSUE #293: `CSRRS`/`CSRRC` to read-only `time` accepted when `rs1 != x0` holds zero
# `CSRRS`/`CSRRC` to read-only `time` accepted when `rs1 != x0` holds zero
## Affected revision
2026-09-06; x86-64 host, RV64 Linux guest user-mode execution.
## Summary
A minimal static guest probe installs a `SIGILL` handler and executes three forms against the read-only `time` CSR:
- `csrrs rd, time, x0` (canonical read-only access, legal)
- `csrrs rd, time, t0` with `t0 = 0` (write-form access, illegal)
- `csrrc rd, time, t0` with `t0 = 0` (write-form access, illegal) On native RV64 Linux and `qemu-riscv64`:
```text
csrrs_x0_trap=0
csrrs_x0_out=<time value>
csrrs_zero_reg_trap=1
csrrc_zero_reg_trap=1
sigill_seen=1
```
Only the canonical `rs1 = x0` read form is accepted; the two forms with `rs1 = t0` (even though `t0` contains zero) trap. On RVVM userland, all three forms are accepted and return the timer value. RVVM appears to use "source value == 0" as a proxy for "source register is x0", which is not correct: write-detection must be based on the register number, not the runtime value.
## Why this is a real defect
The native reference is stable and reproducible; QEMU matches the same semantic shape; the witness is minimal; and this is a CSR decode / write-detection bug rather than a generic accessibility mismatch.
## Existing reports

I searched the public issue tracker and found no existing report for this defect.

carlosqwqqwq:
# csr: validate CSRRS/CSRRC writes by register index

Fixes #293

## Commit message
```text
csr: validate CSRRS/CSRRC writes by register index
```
## Description
The CSR helper decides whether `CSRRS`/`CSRRC` writes from the runtime value loaded from `rs1`. A nonzero `rs1` register containing zero is therefore mistakenly treated as a read, allowing writes to read-only `time`. Carry the architectural write-intent bit from the decoder into the helper: only `rs1=x0`/`zimm=0` forms are read-only accesses.
## Validation
- Native RISC-V hardware and QEMU trap `csrrs`/`csrrc` with `rs1 = t0` (value zero) against read-only `time` and accept only the canonical `rs1 = x0` read form; the patched build matches both.

### DIFF
--- a/src/core/rvvm.c
+++ b/src/core/rvvm.c
@@ -1161,7 +1161,7 @@ PUBLIC rvvm_hart_t* rvvm_create_user_thread(rvvm_machine_t* machine)
 #if defined(USE_FPU)
         // Initialize FPU by writing to status CSR
         rvvm_uxlen_t mstatus = (FS_INITIAL << 13);
-        riscv_csr_op(thread, CSR_MSTATUS, &mstatus, CSR_SETBITS);
+        riscv_csr_op(thread, CSR_MSTATUS, &mstatus, CSR_SETBITS, true);
 #endif
 
 #if defined(USE_JIT)
--- a/src/cpu/riscv_csr.c
+++ b/src/cpu/riscv_csr.c
@@ -815,11 +815,11 @@ static forceinline bool riscv_csr_op_internal(rvvm_hart_t* vm, uint32_t csr_id, 
     return false;
 }
 
-bool riscv_csr_op(rvvm_hart_t* vm, uint32_t csr_id, rvvm_uxlen_t* dest, uint8_t op)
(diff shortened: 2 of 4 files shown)


LekKit:
The `riscv_csr_topei()` helper still uses `riscv_is_csr_write()` instead of the `write` argument. This should be fixed, then `riscv_is_csr_write()` can be dropped as its sole caller is gone. Also worth renaming `write` -> `is_write` because it collides with a POSIX function name otherwise.

PR #304, opened 2026-09-05
riscv: implement fcvtmod.w.d with modulo-2^32 semantics

### LINKED ISSUE #297: `fcvtmod.w.d` uses saturating conversion instead of modulo-2^32 semantics
# `fcvtmod.w.d` uses saturating conversion instead of modulo-2^32 semantics

## Affected revision

2026-09-06; RV64 guest user-mode execution. The interpreter path
(`src/cpu/riscv_fpu.c`, `riscv_emulate_f_opc_op_impl`) and the JIT-assisted path
are both affected.

## Summary

The Zfa `fcvtmod.w.d` instruction must convert its double operand to an integer with **round-towards-zero**, then **take the low 32 bits of the rounded two's-complement result** (modulo 2^32, no saturation). NaN and +/-infinity convert to 0. RVVM instead calls `fpu_fcvt_f64_to_i32`, a **saturating** conversion: any value outside [-2^31, 2^31) produces 0x7fffffff / 0x80000000 (and NaN/infinity saturate instead of yielding 0).

The same opcode also has a separately reported control-flow defect: the handler
writes `rd` and then falls through into the illegal-instruction path because it
lacks a `return`. That issue is tracked independently in #294; the semantic
defect below remains even when the control-flow fix is applied.

## Evidence

Differential probe `repro.S` (28 boundary cases, 4 passes, `fcvtmod.w.d` with `rtz`):

- QEMU reference (`-cpu max`, authoritative Zfa semantics)
- RVVM fixed build: output and flags match QEMU; no process-level trap
- RVVM saturating build (missing-return already fixed, saturating conversion retained): **18 of 27 distinct cases wrong**

Representative mismatches (QEMU vs saturating RVVM):

| input | QEMU result | RVVM (saturating) result |
|---|---|---|
| 2^31 | 0xffffffff80000000 (INT32_MIN) | 0x000000007fffffff |
| 2^32 - 1 | 0xffffffffffffffff (-1) | 0x000000007fffffff |
| 2^63 - 1 | 0xfffffffffffffc00 (-1024) | 0x000000007fffffff |
| +inf | 0 | 0x000000007fffffff |
| -inf | 0 | 0xffffffff80000000 |
| qNaN | 0 | 0x000000007fffffff |
| 2^32 | 0 | 0x000000007fffffff |

The available native RV64 reference does not implement Zfa, so `fcvtmod.w.d` traps there and QEMU `-cpu max` is used as the architectural reference.

## Root cause

In `src/cpu/riscv_fpu.c`, `case 0x08` (fcvtmod.w.d) reuses `fpu_fcvt_f64_to_i32`, whose contract is a saturating convert with NV/inexact handling. The Zfa spec (unpriv, `zfa.adoc`, fcvtmod.w.d) requires RTZ + low-32-bit extraction. The fix adds a dedicated `fpu_fcvtmod_f64_to_i32` in `src/util/fpu_lib.h` implementing the modulo semantics (NV on NaN/inf/overflow, NX on inexact) and calls it from the interpreter. The missing `return` remains the separate scope of #294/#303.

## Fix

See `fix.patch`: adds `fpu_fcvtmod_f64_to_i32` to `fpu_lib.h`; `riscv_fpu.c`
case 0x08 now uses it. The missing `return` is covered separately by #294/#303.

## Validation

- The modulo conversion path matches QEMU `-cpu max` on the result/flag matrix;
  the legal-instruction retirement control is covered separately by #294/#303.
- The saturating variant fails 18/27 cases, proving the defect is in the
  conversion semantics, not the control flow.

## Existing reports

A separate report covers only the missing-return/partial-commit aspect of the same instruction. No existing report covers the saturating-vs-modulo semantic error; the two defects are fixed by different changes (control flow vs conversion routine).

carlosqwqqwq:
# riscv: implement fcvtmod.w.d with modulo-2^32 semantics

## Commit message

```text
riscv: implement fcvtmod.w.d with modulo-2^32 semantics

fcvtmod.w.d (Zfa) converts its double operand with round-towards-zero
and returns the low 32 bits of the rounded two's-complement result
(modulo 2^32, no saturation); NaN and infinity convert to 0.
The previous handler reused the saturating fpu_fcvt_f64_to_i32,
which yields 0x7fffffff/0x80000000 for out-of-range inputs and
saturates NaN/infinity.

Add fpu_fcvtmod_f64_to_i32 implementing the modulo semantics with
NV on NaN/inf/overflow and NX on inexact rounding, and use it from the
fcvtmod.w.d case. The separate architectural writeback control-flow fix is
tracked by PR #303.
```

## Description

`fpu_lib.h`: new `fpu_fcvtmod_f64_to_i32` performs RTZ rounding on the raw double bits, extracts the low 32 bits of the rounded value, and applies the sign; NV is raised for NaN/infinity/out-of-range, NX for inexact truncation. The in-range path uses direct bit-level extraction for the common case (|x| < 2^31).

`riscv_fpu.c`: case 0x08 calls the new routine; reserved rounding-mode
encodings retain the existing illegal path. The separate fall-through defect
is handled by PR #303, and this semantic patch remains independently
applicable to the same `staging` baseline.

## Validation

- The 28-case boundary matrix matches QEMU `-cpu max` for the modulo result and
  `fflags` produced by this conversion path; instruction retirement remains the
  separate scope of PR #303.
- The saturating variant mismatches on 18/27 distinct cases, confirming the
  conversion semantics are independent of the control-flow defect in #303.

Fixes #297

### DIFF
--- a/src/cpu/riscv_fpu.c
+++ b/src/cpu/riscv_fpu.c
@@ -244,7 +244,7 @@ static slow_path void riscv_emulate_f_opc_op_impl(rvvm_hart_t* vm, const uint32_
                         break;
                     case 0x08: // fcvtmod.w.d (Zfa)
                         if (likely(rm == 0x01)) {
-                            riscv_write_reg(vm, rds, (int32_t)fpu_fcvt_f64_to_i32(riscv_view_d(vm, rs1)));
+                            riscv_write_reg(vm, rds, fpu_fcvtmod_f64_to_i32(riscv_view_d(vm, rs1)));
                         }
                         break;
                 }
--- a/src/util/fpu_lib.h
+++ b/src/util/fpu_lib.h
@@ -1477,6 +1477,78 @@ static forceinline int32_t fpu_fcvt_f64_to_i32(fpu_f64_t d)
     return 0x7FFFFFFFU;
 }
 
+static forceinline int32_t fpu_fcvtmod_f64_to_i32(fpu_f64_t d)
+{
+    /*
+     * fcvtmod.w.d (Zfa): convert a double to int32 with round-towards-zero,
+     * returning the low 32 bits of the rounded two's complement result
+     * (modulo 2^32, no saturation). NaN and infinity convert to 0.
+     * Exception flags match fcvt.w.d: NV on NaN/inf/overflow, NX on
+     * inexact rounding.
+     */
+    const uint64_t u    = fpu_bit_f64_to_u64(d);
+    const uint64_t sign = u >> 63;
+    const uint64_t exp  = (u >> 52) & 0x7FF;
+    const uint64_t frac = u & 0xFFFFFFFFFFFFFULL;
+
+    if (unlikely(exp == 0x7FF)) {
+        // NaN or infinity: result is 0, raise NV
+        fpu_raise_invalid();
+        return 0;
+    }
+    if (unlikely(exp == 0)) {
+        // Zero or denormal: |d| < 2^-1022, RTZ gives 0
+        if (unlikely(frac != 0)) {
+            fpu_raise_inexact();
+        }
+        return 0;
+    }
+
+    const int64_t e = (int64_t)exp - 1023;      // unbiased exponent
+    const uint64_t m = frac | (1ULL << 52);     // implicit leading bit
+    uint64_t r;
+    bool overflow = false;
+    bool inexact  = false;
+
+    if (e <= 52) {
+        if (e >= 0) {
+            const int sh = 52 - (int)e;
+            r = m >> sh;
+            if (unlikely(m & ((1ULL << sh) - 1))) {
+                inexact = true;
+            }
+        } else {
+            // |d| < 1.0: RTZ gives 0
+            r = 0;
+            inexact = true;
+        }
+        if (e < 31) {
+            // Result fits in int32
+        } else if (e == 31) {
+            // Only INT32_MIN is exactly representable
+            overflow = !(sign && r == (1ULL << 31));
+        } else {
+            overflow = true;
+        }
+    } else {
+        // |d| >= 2^53: already an integer, may still have bits to return
+        const int sh = (int)e - 52;
+        r = (sh < 64) ? (m << sh) : 0;
+        overflow = true;
+    }
+
+    if (sign) {
+        r = 0 - r;
+    }
+
+    if (unlikely(overflow)) {
+        fpu_raise_invalid();
+    } else if (unlikely(inexact)) {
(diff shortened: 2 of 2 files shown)


src/cpu/riscv_fpu.c
```diff
@@ -244,7 +244,7 @@ static slow_path void riscv_emulate_f_opc_op_impl(rvvm_hart_t* vm, const uint32_
                         break;
                     case 0x08: // fcvtmod.w.d (Zfa)
                         if (likely(rm == 0x01)) {
-                            riscv_write_reg(vm, rds, (int32_t)fpu_fcvt_f64_to_i32(riscv_view_d(vm, rs1)));
+                            riscv_write_reg(vm, rds, fpu_fcvtmod_f64_to_i32(riscv_view_d(vm, rs1)));
```
LekKit, line 247: This now conflicts with an in-tree return on this instruction. The branch should be rebased, keeping the in-tree return after this line.

src/util/fpu_lib.h
```diff
@@ -1477,6 +1477,78 @@ static forceinline int32_t fpu_fcvt_f64_to_i32(fpu_f64_t d)
...
+    if (sign) {
+        r = 0 - r;
+    }
+
+    if (unlikely(overflow)) {
+        fpu_raise_invalid();
+    } else if (unlikely(inexact)) {
+        fpu_raise_inexact();
+    }
+    return (int32_t)r;
```
LekKit: Use this:
```
return (int32_t)(uint32_t)r;
```

Otherwise this is a UB whenever `r` does not fit into 32 bits.

src/util/fpu_lib.h
```diff
@@ -1477,6 +1477,78 @@ static forceinline int32_t fpu_fcvt_f64_to_i32(fpu_f64_t d)
...
+     * fcvtmod.w.d (Zfa): convert a double to int32 with round-towards-zero,
+     * returning the low 32 bits of the rounded two's complement result
+     * (modulo 2^32, no saturation). NaN and infinity convert to 0.
+     * Exception flags match fcvt.w.d: NV on NaN/inf/overflow, NX on
+     * inexact rounding.
+     */
+    const uint64_t u    = fpu_bit_f64_to_u64(d);
+    const uint64_t sign = u >> 63;
+    const uint64_t exp  = (u >> 52) & 0x7FF;
+    const uint64_t frac = u & 0xFFFFFFFFFFFFFULL;
```
LekKit: Use `FPU_LIB_FP64_MANTISSA_MASK` instead of `0xFFFFFFFFFFFFFULL`

PR #315, opened 2026-09-06
riscv: avoid FMA error recovery for directed fmadd.s

### LINKED ISSUE #309: `fmadd.s` directed rounding returns wrong results and underreports UF
# `fmadd.s` directed rounding returns wrong results and underreports UF

## Affected revision

RVVM staging snapshot used for this report, 2026-09-05, x86-64 host, RV64 guest user-mode execution (interpreter path; the x86 JIT falls back to the interpreter for FMA, so both are affected identically).
## Summary

For the directed rounding modes (`rtz`/`rdn`/`rup`), `fmadd.s` returns the wrong result —the f32 neighbor on the wrong side of an exactly representable value —or underreports `UF` on subnormal results. The software FMA fallback of `fpu_fma32` (`fpu_lib.h`, used on hosts without `FPU_LIB_OPTIMAL_BUILTIN_FMA`, i.e. x86-64) computes

```c
sum = fpu_add64(mul, add);            // f64 add under the guest directed mode
err = fpu_add_error64(sum, mul, add); // Knuth TwoSum
ret = fpu_fcvt_f64_to_f32(fpu_odd_round64(sum, err));

```

The TwoSum error term is only exact in round-to-nearest. Under `rtz`/`rdn`/ `rup` it degenerates to a wrong sign, a wrong magnitude, or a spurious `+/-0` (e.g. `err = -0.0` while the true residual is `+2^-70`), and the sign-based ±1-ulp nudge of `fpu_odd_round64` then crosses an f32 grid point, so the second f64->f32 directed conversion rounds to the wrong neighbor. When the result bits happen to survive (the f64 sum is exactly an f32 value), the f32-level `UF` is still missing: the only inexact step is the f64 add, which cannot raise the f32 underflow flag because an f32 subnormal result is a normal f64 operand.
## Inputs (all finite `fmadd.s` operands)

| case | rs1 | rs2 | rs3 |
|---|---|---|---|
| c0 | `0x979e1b59` | `0xc3fecd4d` | `0x76995000` |
| c1 | `0x0bf34dad` | `0x00000001` | `0x00006076` |
| c2 | `0x868fc4b6` | `0x00000001` | `0x000036bf` |
| c3 | `0x880dd88c` | `0x00000001` | `0x8000603d` |
`rs2 = 0x00000001` is the smallest positive subnormal (`+2^-149`), so `c1`-`c3` exercise f32 subnormal results whose exact value needs bits below the f32 subnormal grid; `c0` is an exact-f32-grid-point case with a tiny positive residual.
## Reference behavior

Native RISC-V hardware and `qemu-riscv64` agree on every case and mode, and the expected values were independently verified with exact rational arithmetic:

```text
c0 rne res=76995000 fl=1    c0 rtz res=76995000 fl=1
c0 rdn res=76995000 fl=1    c0 rup res=76995001 fl=1
c1 rne res=00006076 fl=3    c1 rtz res=00006076 fl=3
c1 rdn res=00006076 fl=3    c1 rup res=00006077 fl=3
c2 rne res=000036bf fl=3    c2 rtz res=000036be fl=3
c2 rdn res=000036be fl=3    c2 rup res=000036bf fl=3
c3 rne res=8000603d fl=3    c3 rtz res=8000603d fl=3
c3 rdn res=8000603e fl=3    c3 rup res=8000603d fl=3

```

## RVVM result (unmodified build; interpreter and JIT identical)

```text
c0 rdn res=76994fff fl=1     (should be 76995000; neighbor below)
c1 rdn res=00006075 fl=3     (should be 00006076; neighbor below)
c2 rup res=000036bf fl=1     (result ok, fl should be 3: UF missing)
c3 rtz res=8000603c fl=3     (should be 8000603d; neighbor below)
c3 rup res=8000603d fl=1     (result ok, fl should be 3: UF missing)

```

A 16000-line random differential (4000 cases x 4 modes) was run in addition to the four fixed cases above. In the directed modes, 173 output lines with finite operands differ from native RISC-V hardware/QEMU (53 `rtz` + 76 `rdn` + 44 `rup`, of which 11 are flags-only); `rne` is unaffected for finite operands.
## Root cause

`fpu_add_error64` (Knuth TwoSum, `fpu_lib.h`) is only guaranteed exact under round-to-nearest. Under `rtz`/`rdn`/`rup` the intermediate directed subtractions are themselves rounded, so the returned "error" is garbage:
- `c0 rdn`: true residual `+5.2e-22`, computed `err = -0.0` (`0x8000000000000000`) -> `fpu_odd_round64` sees a nonzero negative error and nudges the f64 sum (which is exactly `0x76995000`) down by one f64 ulp; the second `rdn` f64->f32 conversion then rounds to `0x76994fff`;
- `c1 rdn`: same pattern, result `0x00006075` instead of `0x00006076`;
- `c3 rtz`: garbage negative error nudges down, `0x8000603c` instead of `0x8000603d`;
- `c2`/`c3 rup`: the TwoSum happens to return `+0.0` (no nudge), so the result bits are right, but the exact value is tiny and inexact while the f64 sum is exactly an f32 value: the host f64 add raises `NX` and the f64 ->f32 conversion is exact, so no host op raises the f32 `UF` (`fl=1` instead of `fl=3`).
## Repair

For the directed modes, save the NX raised by the f64 `sum` before running
TwoSum, skip the TwoSum/odd-round result adjustment, and reconstruct the
guest flags from that saved inexactness and the f32 result:

```c
ret = fpu_fcvt_f64_to_f32(sum);
const bool inexact = (sum_exceptions & ~old_exceptions & FPU_LIB_FLAG_NX) != 0;
uint32_t exceptions = fpu_get_exceptions();
exceptions = (exceptions & ~(FPU_LIB_FLAG_NX | FPU_LIB_FLAG_UF)) |
             (old_exceptions & (FPU_LIB_FLAG_NX | FPU_LIB_FLAG_UF));
if (inexact) {
    exceptions |= FPU_LIB_FLAG_NX;
    if ((fpu_bit_f32_to_u32(ret) & FPU_LIB_FP32_NOSIGNED_MASK) < FPU_LIB_FP32_MINIMUM_NORM) {
        exceptions |= FPU_LIB_FLAG_UF;
    }
}
fpu_set_exceptions(exceptions);

```

The f64 sum is the correctly directed f64 rounding of the exact `a*b+c` (the
product of two f32 values is exact in f64), and the saved f64 NX records the
double-rounding loss even when the f64 value is already on an f32 grid point.
The f32 result determines whether that inexact operation is also tiny, so UF
is restored explicitly. The RNE/RMM and special-value paths remain outside this
report. This patch is independently applicable to the `staging` branch.
## Verification

- With the proposed change, the four fixed cases and the directed-mode rows match native RISC-V hardware/QEMU (result + flags).
- The random differential shows zero differences on all finite-operand lines (173 lines fixed); the only residual differences are NaN-payload propagation cases on inf/NaN operands, which a separate NaN-canonicalization patch handles.
- RNE/RMM and inf/NaN cases remain outside this standalone patch's scope.

carlosqwqqwq:
# riscv: avoid FMA error recovery for directed fmadd.s

Fixes #309

## Commit message

```text
riscv: avoid FMA error recovery for directed fmadd.s

```

## Description

The `fpu_fma32` error-recovery calculation is exact only for RNE. For directed modes it can cross an f32 grid point and under-report `UF`. Apply the f32-specific single-rounding path, preserve the NX raised by the preceding f64 sum, and discard only the later TwoSum flags that are not guest-visible. The change is limited to `src/util/fpu_lib.h`; validation artifacts stay out of the source diff.
## Validation

- The witness runs on native RISC-V hardware, QEMU, and the affected RVVM build; the patched build matches both references.

The patch is independently applicable to the `staging` branch. Its scope is
finite `fmadd.s` directed rounding under RTZ/RDN/RUP; other FMA root causes
remain outside this patch.

### DIFF
--- a/src/util/fpu_lib.h
+++ b/src/util/fpu_lib.h
@@ -1070,9 +1070,26 @@ static forceinline func_opt_size fpu_f32_t fpu_fma32(fpu_f32_t a, fpu_f32_t b, f
     fpu_f64_t mul = fpu_mul64(fpu_fcvt_f32_to_f64(a), fpu_fcvt_f32_to_f64(b));
     fpu_f64_t add = fpu_fcvt_f32_to_f64(c);
     fpu_f64_t sum = fpu_add64(mul, add);
+    uint32_t sum_exceptions = fpu_get_exceptions();
     fpu_f64_t err = fpu_add_error64(sum, mul, add);
-    fpu_f64_t res = fpu_odd_round64(sum, err);
-    fpu_f32_t ret = fpu_fcvt_f64_to_f32(res);
+    const uint32_t mode = fpu_get_rounding_mode();
+    fpu_f32_t ret;
+    if (mode == FPU_LIB_ROUND_TZ || mode == FPU_LIB_ROUND_DN || mode == FPU_LIB_ROUND_UP) {
+        ret = fpu_fcvt_f64_to_f32(sum);
+        const bool inexact = (sum_exceptions & ~old_exceptions & FPU_LIB_FLAG_NX) != 0;
+        uint32_t exceptions = fpu_get_exceptions();
+        exceptions = (exceptions & ~(FPU_LIB_FLAG_NX | FPU_LIB_FLAG_UF)) |
+                     (old_exceptions & (FPU_LIB_FLAG_NX | FPU_LIB_FLAG_UF));
+        if (inexact) {
+            exceptions |= FPU_LIB_FLAG_NX;
+            if ((fpu_bit_f32_to_u32(ret) & FPU_LIB_FP32_NOSIGNED_MASK) < FPU_LIB_FP32_MINIMUM_NORM) {
+                exceptions |= FPU_LIB_FLAG_UF;
+            }
+        }
+        fpu_set_exceptions(exceptions);
+    } else {
+        ret = fpu_fcvt_f64_to_f32(fpu_odd_round64(sum, err));
+    }
 #if defined(USE_SOFT_FPU_FENV)
     fpu_soft_fenv_check_add64(fpu_fcvt_f32_to_f64(ret), sum, err);
 #endif


LekKit:
The result is correct, but flags regress: Only NX raised by fadd64 is kept, the NX/UF raised by f64->f32 is discarded. Whenever the f64 sum is exact but not representable in f32, NX is lost.
Fix: don't touch NX at all in the directed path, and only add the UF that the host cannot see (f64 add inexact, conversion exact, f32 result subnormal). Don't compute the TwoSum there either:
```c
if (mode == FPU_LIB_ROUND_TZ || mode == FPU_LIB_ROUND_DN || mode == FPU_LIB_ROUND_UP) {
    ret = fpu_fcvt_f64_to_f32(sum);
    if ((fpu_get_exceptions() & ~old_exceptions & FPU_LIB_FLAG_NX)
        && (fpu_bit_f32_to_u32(ret) & FPU_LIB_FP32_NOSIGNED_MASK) < FPU_LIB_FP32_MINIMUM_NORM) {
        fpu_raise_exceptions(FPU_LIB_FLAG_UF); // Inexact and tiny after rounding
    }
} else {
    err = fpu_add_error64(sum, mul, add);
    ret = fpu_fcvt_f64_to_f32(fpu_odd_round64(sum, err));
}
```

PR #317, opened 2026-09-12
Machine snapshots: serialize RAM, harts and devices

zmarlon:
The snapshot container and the device suspend callback were already in
  place, but nothing walked a machine to drive them, so snapshots couldn't
  actually be taken. 
  
  rvvm_machine_snapshot() serializes RAM, the harts and every device, in
  whichever direction the snapshot was opened. RAM size, hart count, XLEN and
  the timebase are stored and checked, so a snapshot only resumes into a
  machine it fits. TLB and compiled blocks are dropped on resume since 
  they're just caches over guest memory.
  
  Legacy MMIO devices gained the suspend callback the region API already had,
  so the existing implementations take part. I added suspend for three more
  that a typical machine needs: PCI config state including BAR placement,
  NVMe queues, and the OpenCores I2C controller. -loadsnap and -savesnap make
  it usable from the CLI.
  
  Tested by booting a Linux guest with a framebuffer, NVMe disks and PCIe,
  saving it, rebuilding an identical machine and restoring: the guest carries
  on from where it left off.

### DIFF
--- a/include/rvvm/rvvm.h
+++ b/include/rvvm/rvvm.h
@@ -184,11 +184,8 @@ typedef struct {
     /** Called on machine reset */
     void (*reset)(rvvm_mmio_dev_t* dev);
 
-    /*
-     * TODO
-     * void (*suspend)(rvvm_mmio_dev_t* dev, rvvm_state_t* state);
-     * void (*resume)(rvvm_mmio_dev_t* dev, rvvm_state_t* state);
-     */
+    /** Called to serialize/deserialize device state on a paused machine */
+    void (*suspend)(rvvm_mmio_dev_t* dev, rvvm_snapshot_t* snap, bool resume);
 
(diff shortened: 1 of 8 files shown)


LekKit:
All good in general, a few things i would prefer (I may do myself after merging if needed):
- Use a single `-snapshot vm.state` argument both to load and save snapshot (Load on launch, save on exit)
- Delete the old snapshot on load, because in case of a crash, guest disk / snapshot state will very likely be desynced and lead to disk image corruption if anyone tries to load the snapshot again. Snapshot for both state / disk will require separate work on disk subsystem, and a new disk image format
- Move the hunk from `rvvm_snapshot.c` to `rvvm.c` for now. The `rvvm_snapshot` source purely holds the snapshot format implementation, and should not be bothered with CPU / ram / etc serialization details.

